PR #78, opened 2020-09-01
Updating ECS version to 1.6.0

jonathan-buttner:
This PR pulls in ECS version 1.6.0.

Notable changes:

- Short descriptions were added to fields that had a description longer than 120ish characters
- `custom_process.yml` changed a bit since ECS core no longer defines `process.parent` fields, instead it duplicates the `process` fields onto itself as `parent`
- Not longer using my fork for ECS, we're back to ECS `master` branch!

The only things that should have changed from this PR are the descriptions/short fields and some of the metadata from the `process` changes.

### DIFF
--- a/custom_schemas/custom_process.yml
+++ b/custom_schemas/custom_process.yml
@@ -143,11 +143,11 @@
       level: custom
       type: object
       description: >
-        The field set containing parent process info in case of any ppid spoofing.
+        The field set containing process info in case of any pid spoofing. This is mainly useful for process.parent.
 
     - name: Ext.real.pid
       level: custom
       type: long
+      short: The real pid of the process if ppid spoofing is happening.
       description: >
-        The ppid of the process that actually spawned the current process, in case of
-        ppid spoofing.
+        For process.parent this will be the ppid of the process that actually spawned the current process.
--- a/custom_schemas/custom_process.yml
+++ b/custom_schemas/custom_process.yml
@@ -13,6 +13,7 @@
     top_level: true
     expected:
       - Target
+      - { at: Target.process, as: parent }
   type: group
   fields:
     - name: Ext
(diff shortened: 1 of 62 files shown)


custom_schemas/custom_process.yml
```diff
@@ -13,6 +13,7 @@
     top_level: true
     expected:
       - Target
+      - { at: Target.process, as: parent }
```
jonathan-buttner, line 16: The way schema is reused changed in ECS 1.6: https://github.com/elastic/ecs/pull/960
https://github.com/elastic/ecs/issues/959

This is basically placing the combined process schema onto `Target.process.parent`

custom_schemas/custom_process.yml
```diff
@@ -138,76 +139,15 @@
         Leave unpopulated if the validity or trust of the certificate was unchecked.
       example: ERROR_UNTRUSTED_ROOT
 
-    - name: parent
```
jonathan-buttner, line 141: Since the `parent` schema is no longer explicitly specified in the ECS core schema (instead it just duplicates the `process` fields and calls them `process.parent`), we will move these fields onto `process` instead so they can also be duplicated. The fields on `process` won't actually be used, they'll only be used for `process.parent`.

generated/alerts/elasticsearch/7/template.json
```diff
@@ -4,7 +4,7 @@
   ],
   "mappings": {
     "_meta": {
-      "version": "1.5.0"
+      "version": "1.6.0"
```
jonathan-buttner, line 7: The actual mapping shouldn't really change. We're just bumping the version number of ECS and a couple `type: object` fields will be removed (marking it as `object` is optional because that's the default type).

michaelolo24 (approved):

PR #80, opened 2020-09-16
[SECURITY_SOLUTION] remove transform in case we need to revert

kevinlog:
Removing the transform file in case we need to revert for a bit due to this bug: https://github.com/elastic/kibana/issues/77133

There is an issue where transforms aren't starting correctly in cloud deployments.  This will enable testing to continue until we fix the root cause.


nnamdifrankie (approved):

jfsiii:
I believe if we push this (as 0.16.0-dev.1 ?) it will also fix the deploys that are currently broken. 

@nnamdifrankie @jonathan-buttner What's required to get this in https://github.com/elastic/package-storage/tree/snapshot/packages/endpoint ? I can PR but I saw reference to a script so I assume there's a process. LMK if I can help

PR #86, opened 2020-09-22
78024: move transform add template for index

nnamdifrankie:
https://github.com/elastic/kibana/issues/78024

- [x] moved the transform to elasticsearch
- [x] create index_template
- [x] delete old metadata current


package/endpoint/elasticsearch/index_template/metrics-metadata-current.json
```diff
@@ -0,0 +1,255 @@
...
+      "index": {
+        "lifecycle": {
+          "name": "metrics-default"
+        },
+        "codec": "best_compression",
+        "refresh_interval": "5s",
+        "number_of_shards": "1",
+        "query": {
+          "default_field": [
+            "message"
```
jonathan-buttner, line 247: I don't think we want this as the default field right? That field isn't mapped in this template. We can probably just remove the `query` section.

jonathan-buttner (approved):
👍  pending the removal of the `settings.query.default_field` section

package/endpoint/elasticsearch/index_template/metrics-metadata-current.json
```diff
@@ -0,0 +1,250 @@
+{
+    "index_patterns": [
+        "metrics-endpoint.metadata_current-*"
```
ruflin, line 3: Is this the target index for the transform data? We must use something here which does not match the indexing strategy, for example `"metrics-endpoint.metadata_current_*"` could work. It is important that it does NOT match `metrics-*-*`

I know, we set the priority higher below and is the reason this might still work even though it conflicts. But I think it is misleading that some indices are not datastreams.

package/endpoint/elasticsearch/index_template/metrics-metadata-current.json
```diff
@@ -0,0 +1,250 @@
...
+                                    "type": "keyword",
+                                    "ignore_above": 1024
+                                },
+                                "version": {
+                                    "type": "keyword",
+                                    "ignore_above": 1024
+                                }
+                            }
+                        },
+                        "data_stream": {
```
ruflin, line 79: I assume this is not needed as it is not a data_stream anymore? Or what would be the values here?

nnamdifrankie: Oh, this is will be in the document from metadata index. Recall that we are embedding the who latest document inside the document for this index.

ruflin: Got it!

package/endpoint/elasticsearch/index_template/metrics-metadata-current.json
nnamdifrankie: Sounds good

package/endpoint/elasticsearch/index_template/metrics-metadata-current.json
```diff
@@ -0,0 +1,247 @@
...
+                        }
+                    }
+                }
+            ],
+            "date_detection": false,
+            "properties": {
+                "@timestamp": {
+                    "type": "date"
+                },
+                "HostDetails": {
```
ruflin, line 26: Not related to this change but I stumbled over the CamelCase naming here as often work with `_` instead for all the other fields. In case users query this data, it will seem a bit incosistent.

ruflin (approved):

package/endpoint/elasticsearch/index_template/metrics-metadata-current.json
nnamdifrankie: From previous conversations I have had with the team, we were ask to make custom element capitalized hence the casing.

nnamdifrankie:
https://github.com/elastic/security-team/issues/77

Issue #83, opened 2020-09-18
Parse DNS Fields


jonathan-buttner:
Create a pipeline that adds the following fields:

![image](https://user-images.githubusercontent.com/56361221/93636309-b2ae6780-f9c1-11ea-957b-5930487cc563.png)

These fields are already populated by the elastic endgame endpoint.

Take a look at these docs for writing the ingest pipeline to do the parsing: https://www.elastic.co/guide/en/elasticsearch/reference/current/grok-processor.html

Related:
https://github.com/elastic/endpoint-dev/issues/7136

james-elastic:
AC: A user will be able to see the three fields from the screenshot.

jonathan-buttner:
PR is here: https://github.com/elastic/endpoint-package/pull/88

It populates `dns.question.type` and `dns.question.name`. I looked into populating `dns.question.registered_domain` but it'd be pretty complicated to get it right using painless scripting I think according to the ecs spec: https://github.com/elastic/ecs/blob/master/schemas/dns.yml#L102

Also `dns.question.name` doesn't full adhere to the ecs spec with regards to having non-printable characters: https://github.com/elastic/ecs/blob/master/schemas/dns.yml#L74

jonathan-buttner:
Closing this ticket as the PR above was merged.

PR #88, opened 2020-09-30
Populate quesiton.type for dns events

jonathan-buttner:
This PR adds a couple pipeline processors to parse the DNS network events `message` field. It uses the `message` field to populate `dns.question.type` and `dns.question.name`. I also switched the pipeline over to be a yml file instead of json so I could write a multiline script.

@ferullo @jrmolin once this goes in we'll need to coordinate on any changes to the DNS events `message` field. Also could you point me at the endgame sensor's `question.type` parsing code for number -> string representation. I just want to make sure I miss any.

I'm using this regex:

```
^DNS query is completed for the name .* type %{WORD:dns.question.Ext.type}
```

`WORD` is defined here: https://github.com/elastic/elasticsearch/blob/7.9/libs/grok/src/main/resources/patterns/grok-patterns#L14

Since it's possible the domain name could have invalid characters in the request I figured it's best to get everything and assume a comma ends the field. Let me know if you have better ideas though.

Reference:
https://github.com/elastic/endpoint-package/issues/83

<details><summary>Working example</summary>

image

</details>

cc: @MikePaquette


marshallmain:
Out of curiosity why do we need to populate these fields in the pipeline rather than have the endpoint populate the fields when it builds the rest of the event?

jonathan-buttner:
> Out of curiosity why do we need to populate these fields in the pipeline rather than have the endpoint populate the fields when it builds the rest of the event?

Yeah @jrmolin or @ferullo can probably answer better here. I believe certain fields are more difficult than others to parse on the endpoint side. I think most of the information was retrieved in the ETW events for the original sensor and they have to track multiple events to get the information. But since we're already putting the information in the `message` field, I'm not sure why we couldn't just populate the `dns.question.type` and `dns.quesiton.name` fields?

ferullo:
@jonathan-buttner if/when Endpoint begins to populate this data will the pipeline still overwrite with Endpoint sets or will this code be effectively disabled?

jonathan-buttner:
> @jonathan-buttner if/when Endpoint begins to populate this data will the pipeline still overwrite with Endpoint sets or will this code be effectively disabled?

The way this PR is right now, it'd overwrite what the endpoint puts in those fields. I can add a simple `if` to check if the fields are already present though and skip populating them.

ferullo:
I think that would be a good check to add. Marshall makes a good point about Endpoint doing this itself. The reality though is Endpoint won't get that done for 7.10 and functionally it doesn't matter if Endpoint or the pipeline does this until detection rules can be run in Endpoint. We should move this logic to Endpoint when that happens but if there's no reason to not merge this it seems like it'll give detection rules the data they need in the meantime.

Does that sound reasonable?

marshallmain:
Once we add and ship this pipeline we can't really remove it from the package unless we can ensure that all endpoints that use it have been upgraded to populate the fields on the endpoint side, or we'd have to be comfortable breaking compatibility at some point between new package versions and old endpoints. So the risk is that long term we build up a bunch of these fixes applying subtly different modifications and end up with a complex pipeline if we're not careful about what we use it for.

I think the `if` check would be a good addition, and since we're only adding fields that don't exist in the document then this is a fine approach to make the `dns` fields available in 7.10. If we can stick to only adding fields that don't already exist with the pipeline and ensure that we don't modify existing fields then the pipeline shouldn't get too complex even if we add more in the future.

jonathan-buttner:
> Once we add and ship this pipeline we can't really remove it from the package unless we can ensure that all endpoints that use it have been upgraded to populate the fields on the endpoint side, or we'd have to be comfortable breaking compatibility at some point between new package versions and old endpoints. So the risk is that long term we build up a bunch of these fixes applying subtly different modifications and end up with a complex pipeline if we're not careful about what we use it for.
> 
> I think the `if` check would be a good addition, and since we're only adding fields that don't exist in the document then this is a fine approach to make the `dns` fields available in 7.10. If we can stick to only adding fields that don't already exist with the pipeline and ensure that we don't modify existing fields then the pipeline shouldn't get too complex even if we add more in the future.

Yeah that's a good point. I agree, hopefully we can stick to adding fields that don't exist in the pipelines and avoid modifying existing fields. If we end up having to make breaking changes we should wait until an 8.x release for that.

I'll add the check to see if the field already exists.

package/endpoint/dataset/network/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,71 @@
...
+      ignore_failure: true
+      field: "message"
+      pattern_definitions:
+        DOMAIN_NAME: "[^,]+"
+      patterns:
+        # dns.question.Ext.type is a temporary field used to hold the parsed question type
+        - "^DNS query is completed for the name %{DOMAIN_NAME:dns.question.name}, type %{WORD:dns.question.Ext.type}"
+  - script:
+      ignore_failure: true
+      if: "ctx.network.protocol == 'dns' && ctx.dns?.question?.Ext?.type != null && ctx.dns?.question?.type == null"
```
jonathan-buttner, line 44: If above `grok` doesn't run then this processor won't run either because it needs the `dns.quesiton.Ext.type` to be present.

package/endpoint/dataset/network/elasticsearch/ingest_pipeline/default.yml
```diff
@@ -0,0 +1,71 @@
...
+        - continent_name
+        - country_name
+        - country_iso_code
+        - region_iso_code
+        - region_name
+        - city_name
+        - location
+      ignore_missing: true
+  - grok:
+      if: "ctx.network.protocol == 'dns' && ctx.dns?.question?.name == null"
```
jonathan-buttner, line 33: if `dns.question.type` exists but the `name` doesn't, `name` would still be filled in using `grok` here.

jonathan-buttner:
I must have completely overlooked this but the endpoint is already populating `dns.question.name` 😅 . I'll switch this PR to just populate `dns.question.type`.

Doesn't look like linux or mac send dns events, but I assume once they do they can just populate the `dns.question.name` as well.

michaelolo24 (approved):

PR #101, opened 2020-10-29
Align alerts and events mapping

jonathan-buttner:
This PR tries to align the mapping for alerts and events. I split this up into 2 commits, one that just has the subset changes and the other has the generated files. Hopefully that makes it a bit easier to review.

I did not change `metadata` because some additional things have to happen to generate the mappings for the transform that is used. I'll do that in a follow up PR.

_**NOTE:**_ There shouldn't be any _**deletions**_ in the generated files.

The general way to review this is look at the `malware_event.yml` https://github.com/elastic/endpoint-package/blob/master/custom_subsets/elastic_endpoint/alerts/malware_event.yaml file and make sure the `process`, `host`, `user`, `group`, and `file` fields in the other directories (file, library, process etc) are pretty close. I didn't copy over all the `Ext` fields because most of them will only apply to the alerts. But most of the core ecs fields should be there for alerts and events. 

Things that changed:
- I tried to organize the fields in alphabetical order for the things that I touched
- More `host` fields
- More `pe` fields
- More `file` fields
- More `hash` fields
- More `events` fields
- More `code_signature` fields

Long term we might want to implement way better way to keep these in sync, like some sort of _include_ functionality or something.

### DIFF
--- a/custom_subsets/elastic_endpoint/process/process.yaml
+++ b/custom_subsets/elastic_endpoint/process/process.yaml
@@ -21,20 +21,23 @@ fields:
         fields: "*"
   host:
     fields:
-      name: {}
+      architecture: {}
+      domain: {}
       hostname: {}
+      id: {}
       ip: {}
       mac: {}
-      architecture: {}
-      id: {}
+      name: {}
+      type: {}
+      uptime: {}
       os:
         fields:
+          family: {}
+          full: {}
+          kernel: {}
           platform: {}
           version: {}
-          family: {}
           name: {}
-          kernel: {}
-          full: {}
           Ext:
             fields:
               variant: {}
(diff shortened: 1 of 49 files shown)


custom_subsets/elastic_endpoint/process/process.yaml
```diff
@@ -78,49 +101,111 @@ fields:
...
     fields:
-      entity_id: {}
-      pe:
-        fields:
-          original_file_name: {}
-      pid: {}
-      name: {}
-      executable: {}
       args: {}
+      args_count: {}
```
jonathan-buttner, line 105: Just pointing this one out as something the protections team needed I think.

michaelolo24 (approved):

marshallmain (approved):

PR #105, opened 2020-10-30
Update metadata schema

jonathan-buttner:
This PR brings the metadata schema in sync with the other events indices. It's not super necessary but I figured it'd be nice to keep things in sync. My plan is to deploy this in the 7.11 release.

It mainly just adds a few more mapped fields for `events.*` and `host.*`. I also tried to alphabetize the subset file.

There shouldn't be any deletions of fields just additions.


package/endpoint/elasticsearch/index_template/metrics-metadata-current.json
```diff
@@ -112,114 +112,156 @@
...
-                                            "type": "keyword",
-                                            "ignore_above": 1024
+                                            "ignore_above": 1024,
+                                            "type": "keyword"
                                         },
                                         "full": {
-                                            "type": "keyword",
-                                            "ignore_above": 1024,
                                             "fields": {
+                                                "caseless": {
```
jonathan-buttner, line 217: I don't think these are required but these `caseless` fields exist in the metadata data_stream

image


So I figured I'd include them here as well. I just copied portions of the ECS generated template file for these changes.

jonathan-buttner: Searching using the new fields works:

image

nnamdifrankie (approved):
Thank you for the test.

PR #118, opened 2020-12-09
add collection data stream and ilm policy

nnamdifrankie:
This PR adds the new `data stream` that will be used for diagnostic alerts.

It contains:
- The `hidden` field which will make the diagnostic `data stream` hidden
- A custom `ilm_policy` which will be aggressive

These fields are now supported in the package-spec:
https://github.com/elastic/package-spec/pull/91
https://github.com/elastic/package-spec/pull/90

And released via the package-registry:
https://github.com/elastic/package-registry/releases/tag/v0.15.0

It's been tested locally via new changes in Fleet package installation.
https://github.com/elastic/kibana/pull/85492
https://github.com/elastic/kibana/pull/85703


package/endpoint/elasticsearch/ilm/diagnotics.json
```diff
@@ -0,0 +1,15 @@
+{
```
kevinlog, line 1: What's the difference between this `ilm_policy` and this one https://github.com/elastic/endpoint-package/pull/118/files#diff-4dd7edc183ee15694b9a686b280db66ea3864789768bb95daa6c4d56140a49d5R1

nnamdifrankie: Ah this should be deleted

package/endpoint/data_stream/collection/elasticsearch/ilm/diagnotics.json
```diff
@@ -0,0 +1,15 @@
...
+                "actions": {
+                    "rollover": {
+                        "max_size": "10gb",
+                        "max_age": "1d"
+                    }
+                }
+            }
+        }
+    }
+}
```
tsg, line 15: I think we need a `delete` phase, otherwise the indices will be rolled over but not actually deleted. In my testing, I did something like this:

```
PUT /_ilm/policy/endpoint.diagnostic.collection
{
  "policy": {
    "phases": {
      "hot": {
        "min_age": "0ms",
        "actions": {
          "rollover": {
            "max_size": "1gb",
            "max_age": "7d",
            "max_docs": 10000
          }
        }
      },
      "delete": {
        "min_age": "1m",
        "actions": {}
      }
    }
  }
}
```

That should mean that the rolled-over index is deleted 1m after it's rotated. Comparing to your version, I also have a limitation of `max_docs`. Not sure we need it, but maybe good for testing.

kevinlog: thanks @tsg I updated the policy

custom_subsets/elastic_endpoint/collection/collection.yaml
```diff
@@ -0,0 +1,28 @@
+---
```
jonathan-buttner, line 1: Just want to double check that this is the intent but this mapping is pretty reduced. The only fields that will be searchable within ES are `@timestamp, data_stream, ecs.version, and the event.*` fields. Is that what we want?

jonathan-buttner (approved):
Just want to check, if for some reason the kibana side changes don't get merged before feature freeze can we still release these changes? Or are we hoping to merge the kibana changes and this PR at the same time?

custom_subsets/elastic_endpoint/collection/collection.yaml
kevinlog: @jonathan-buttner this is the desired functionality.

This will be read by a telemetry server and it will search by `event.ingested`.  We don't need to index any other fields as we'll only need to search for the latest docs.

jonathan-buttner: Sounds good!

kevinlog:
@jonathan-buttner 

> Just want to check, if for some reason the kibana side changes don't get merged before feature freeze can we still release these changes? Or are we hoping to merge the kibana changes and this PR at the same time?

The Kibana side changes are ready, the PRs are up.
https://github.com/elastic/kibana/pull/85703
https://github.com/elastic/kibana/pull/85492

I've tested everything together locally.  My plan is to merge the Fleet changes as soon as I get approval and then I'll merge this

PR #119, opened 2020-12-11
update endpoint-package

kevinlog:
Update endpoint-package to the latest elastic-package so that the linter will work/accept the new fields.


kevinlog:
@jonathan-buttner 

I ran these two commands:
`> go get -u github.com/elastic/elastic-package`
`> go mod vendor `

As you can see it updated a TON of files.  I can confirm that the linter works correctly, but not sure if I did something wrong.  Just let me know if I should revert and approach it differently.

jonathan-buttner:
It should be fine. It just updated all the libraries used in the repo. Can you try running:

```
go mod tidy
go mod vendor
```

That might reduce it a bit. Either way it looks good though 👍

jonathan-buttner (approved):

PR #120, opened 2020-12-15
add antivirus registration section in policy response

kevinlog:
Adds a new section in the Policy response for the register as AV feature

### DIFF
--- a/custom_schemas/custom_endpoint.yml
+++ b/custom_schemas/custom_endpoint.yml
@@ -111,6 +111,24 @@
         not a simple sum of the actions
       short: the overall status of logging
 
+    - name: policy.applied.configurations.antivirus_registration
+      level: custom
+      type: object
+      description: overall antivirus registration configuration and status of the applied policy
+
+    - name: policy.applied.configurations.antivirus_registration.concerned_actions
+      level: custom
+      type: keyword
+      description: all actions that were taken for antivirus registration
+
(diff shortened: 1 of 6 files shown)


custom_schemas/custom_endpoint.yml
```diff
@@ -111,6 +111,24 @@
         not a simple sum of the actions
       short: the overall status of logging
 
+    - name: policy.applied.configurations.antivirus_registration
```
jonathan-buttner, line 114: Do we need to be able to search for the fields in this object? Or is it to denote what the endpoint is sending back in the policy?

If we don't expect to be performing searches on them, then let's mark them as disabled, by adding a `enabled: false` like here: https://github.com/elastic/endpoint-package/blob/12272fe08d1e239c50790fa059012449e2f81bc5/custom_schemas/custom_endpoint.yml#L171

jonathan-buttner (approved):

custom_schemas/custom_endpoint.yml
kevinlog: @jonathan-buttner this is just to denote what the Endpoint is sending back.  We won't need to search for these fields.  I'll make the change.

kevinlog: change is pushed - https://github.com/elastic/endpoint-package/pull/120/commits/4e410c2bf709101bd6aaaf37fbf4d747b1b2d937

PR #122, opened 2020-12-21
expand transform index search to account for namespaces

kevinlog:
Adjust the transform so that it will also search for indices that may have something other than the `default` namespace.

See this bug:
https://github.com/elastic/kibana/issues/86237

### DIFF
--- a/package/endpoint/elasticsearch/index_template/metrics-metadata-current.json
+++ b/package/endpoint/elasticsearch/index_template/metrics-metadata-current.json
@@ -82,7 +82,7 @@
                                     "type": "constant_keyword"
                                 },
                                 "namespace": {
-                                    "type": "constant_keyword"
+                                    "type": "keyword"
                                 },
                                 "type": {
                                     "type": "constant_keyword"
--- a/package/endpoint/elasticsearch/transform/metadata_current/default.json
+++ b/package/endpoint/elasticsearch/transform/metadata_current/default.json
@@ -1,6 +1,6 @@
 {
     "source": {
-        "index": "metrics-endpoint.metadata-default*"
+        "index": "metrics-endpoint.metadata-*"
     },
     "dest": {
         "index": "metrics-endpoint.metadata_current_default"


jonathan-buttner (approved):

package/endpoint/elasticsearch/transform/metadata_current/default.json
```diff
@@ -1,6 +1,6 @@
 {
     "source": {
-        "index": "metrics-endpoint.metadata-default*"
+        "index": "metrics-endpoint.metadata-*"
     },
     "dest": {
         "index": "metrics-endpoint.metadata_current_default"
```
jonathan-buttner, line 6: Thinking about this more, will this cause overlap between multiple namespaces? I think this will place all the transformed events in `metrics-endpoint.metadata_current_default`. Not sure if that's going to cause issues with counts or permissions?

kevinlog: @jonathan-buttner this will cause overlaps with multiple namespaces, but IMO that's OK because we want the admin tab to display all enrolled endpoints, agnostic of spaces, etc.

If users want to use the name-spaced `metrics-endpoint.metadata-*` indices for another purpose, they will still be there.  We'll just have the transform compile them all into one place for our app page.

jonathan-buttner: Sounds good 👍

package/endpoint/elasticsearch/index_template/metrics-metadata-current.json
```diff
@@ -82,7 +82,7 @@
                                     "type": "constant_keyword"
                                 },
                                 "namespace": {
-                                    "type": "constant_keyword"
+                                    "type": "keyword"
```
kevinlog, line 85: @jonathan-buttner in further testing, I realized that I needed to update this mapping to not be `constant` since this will allow documents from multiple `namespaces`.  If this remains `constant_keyword` then the transform will fail because it finds multiple values in documents for a constant field.  Do you see any issues with updating this mapping?

Here's some logs from ES to illustrate the transform failure without it.

```
   │      Caused by: org.elasticsearch.xpack.transform.transforms.TransformException: Destination index mappings are incompatible with the transform configuration.
   │      	at org.elasticsearch.xpack.transform.transforms.ClientTransformIndexer.decorateBulkIndexException(ClientTransformIndexer.java:390) ~[?:?]
   │      	at org.elasticsearch.xpack.transform.transforms.ClientTransformIndexer.lambda$doNextBulk$0(ClientTransformIndexer.java:168) ~[?:?]
   │      	... 55 more
   │      Caused by: org.elasticsearch.index.mapper.MapperParsingException: failed to parse field [HostDetails.data_stream.namespace] of type [constant_keyword] in document with id 'NDNph60k6tpXkzhtL7f1ryQAAAAAAAAA'. Preview of field's value: 'default'
```

kevinlog: I should add that the alternative to this is to just have the Endpoint hardcode the `default` namespace for the `metadata` index only.

jonathan-buttner: hmm, I'm wondering if we should remove all the `dataset` fields from this mapping. There's a couple concerns I have:

We should test upgrading from 7.10. I think the `PUT /_index_template` will probably succeed:

https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-put-template.html
https://github.com/elastic/kibana/blob/master/x-pack/plugins/fleet/server/services/epm/elasticsearch/template/install.ts#L91

What I'm not sure about is if a previous destination index exists `metrics-endpoint.metadata_current_default` it will have the the `dataset.namespace` defined as `constant_keyword`. This this is not a data_stream (it's a plan old index) and fleet doesn't handle rolling over the index because of the mapping change. So I'm wonder if it will still fail even after we change the mapping because it doesn't need to create a new index.

When I look at the SIEM kibana it looks like it has two metadata indices:

image

One difference I see is in the transform version number:

```
"_meta": {
      "_transform": {
        "transform": "metrics-endpoint.metadata-current-default-0.16.0-dev.0",
```

Where the other index has:

```
"_meta": {
      "created_by": "transform",
      "_transform": {
        "transform": "endpoint.metadata_current-default-0.17.0",
```

So we might get lucky in that because we're bumping the endpoint package version from `0.16.x` to `0.17.x` the transform name will change and we might automatically get a new `metrics-endpoint.metadata_current_default`.

Even if we do remove the `dataset` fields from the new mapping we might need to modify the old `metrics-endpoint.metadata_current_default` mapping.

I'm wondering if we should tack on a version number to `metrics-endpoint.metadata_current_default` or maybe move it to an actual data_stream.

@ruflin what do you think?

jonathan-buttner: Actually looks like the fleet transform code should delete the old destination index:
https://github.com/elastic/kibana/blob/master/x-pack/plugins/fleet/server/services/epm/elasticsearch/transform/remove.ts#L58 so we might be ok on upgrade then.

ruflin: If this is an index and it is not following the new indexing strategy, agree probably the `data_stream.*` fields should not exist? But not sure I follow the full issue.

pzl: Ok, just catching up to this.

re: transform & mapping -- I tested a large schema change on package upgrade, and indeed, the transform upgrade code handles mapping changes fine. At the time I didn't know it performed a delete & recreate, but either way it changes the mapping ok.

re: data_stream fields -- These are just copied from the source document. 

```json
"data_stream": {
  "dataset": "endpoint.metadata",
  "namespace": "default",
  "type": "metrics"
},
```

I don't know if this is actually sent _from_ the Endpoint binary itself on it's way to ES. Or that once it hits the data_stream alias name, ES sets those fields (given that Endpoint is the thing that knows to-which namespace it belongs, perhaps this is an Endpoint-sent field).

The transform is just copying the source doc to the destination. That happens to include these. I can see this being useful, if somebody _did_ want to use `HostDetails.data_stream.namespace == "whatever"` as a KQL filter.

From https://www.elastic.co/guide/en/elasticsearch/reference/master/keyword.html#constant-keyword-field-type : 

> Constant keyword is a specialization of the keyword field for the case that all documents in the index have the same value.

Since namespace is actually variable value (though rarely used internally), `keyword` seems to be the correct change here.

pzl (approved):
I think this changeset is correct as-is.

We need to do some validation of it, but expanding the sources to pull from all namespaces is right (and matches the business ask, of having _all_ at the transform destination).

And the `data_stream.namespace` change is correct too. Having it be `constant_keyword` was perhaps an oversight, since namespaces is an under-tested feature (from our team).

We do need to address the merge conflict, however. I believe this change wants to be rolled out _before_ the change that was recently sent into master

perhaps a revert is in order?

cc @jonathan-buttner  on the last bit

package/endpoint/elasticsearch/index_template/metrics-metadata-current.json
jonathan-buttner: > I don't know if this is actually sent from the Endpoint binary itself on it's way to ES. Or that once it hits the data_stream alias name, ES sets those fields (given that Endpoint is the thing that knows to-which namespace it belongs, perhaps this is an Endpoint-sent field).

The endpoint binary does send them. They have to match the index naming structure when it is a data stream though (i.e. `<type>-<dataset>-<namespace>`).

jonathan-buttner:
> And the `data_stream.namespace` change is correct too. Having it be `constant_keyword` was perhaps an oversight, since namespaces is an under-tested feature (from our team).

For data streams the `namespace` field is always a `constant_keyword` because for a single data stream there should only be a single value in that field. That field has some performance benefits.

> We do need to address the merge conflict, however. I believe this change wants to be rolled out _before_ the change that was recently sent into master
> 
> perhaps a revert is in order?
> 
> cc @jonathan-buttner on the last bit

Which stack version are we targeting for this fix? Is it 7.12 or 7.11? If it is 7.12 we can just merge this to master but it wouldn't get released until February. If it is 7.11 here's the flow:

1. Merge to master
2. Backport to 7.11 branch (this repo uses the same backport script as kibana)
3. Release a new 7.11 version (0.17.1) of the endpoint package

kevinlog:
@jonathan-buttner 

> Which stack version are we targeting for this fix? Is it 7.12 or 7.11?

I was trying to target for `7.11.1` - is possible to tag for a patch release?

My reasoning is that I don't want to rush testing this for `7.11.0` being that we're so close, but would be good to release the package in a timely fashion to fix the namespacing issue.

Let me know what you're comfortable with @pzl @jonathan-buttner - pushing to `7.12` isn't out of the question

jonathan-buttner:
> @jonathan-buttner
> 
> > Which stack version are we targeting for this fix? Is it 7.12 or 7.11?
> 
> I was trying to target for `7.11.1` - is possible to tag for a patch release?

Releasing the package to be ready for 7.11.1 is fine. We can just wait until after 7.11.0 is official released. Would you want the fix to apply to both kibana 7.11.0 and 7.11.1? or only 7.11.1?

> 
> My reasoning is that I don't want to rush testing this for `7.11.0` being that we're so close, but would be good to release the package in a timely fashion to fix the namespacing issue.

Yeah that makes sense.

pzl:
Ah, right, we can just cherry-pick this change off master into the version branch we need. That's perfect.

sounds like we're settled on the intent of the rest of the versioning and release, 

so final note on `namespace` type here:

In an actual data_stream index, yeah it would be a constant. Because only one namespace is flowing into one named index. Here, we have an aggregate across namespaces, so it will be a dynamic `keyword`. This index is not a data_stream itself, just a normal index that _happens_ to have fields coming from various data streams. 

in normal elasticsearch, from scratch, you could just 

```http
POST /some-index-im-making-on-the-fly/_doc
{
  "temp": 56,
  "data_stream": {
    "namespace": "thermostat"
  }
}
```

this will just auto-create the `some-index-im-making-on-the-fly` index, and deduce a mapping from this. Perhaps this is from a smart thermostat that uses fields called "data_stream" and "namespace". Won't create an Elasticsearch data stream. Just treated like a normal data field. The names are coincidental, but don't clash or do anything magic.

I think the transform destination index falls into the same category. Its _coming from_ data streams, so it has those field names, but it's being put into just a normal index tracking those data_stream fields as standard data.

pzl:
Circling back with some verification before merging:

To show the before/after I have created two packages locally. The latter contains the namespace change

- 0.18.0-dev.0
- 0.18.0-dev.1 (**contains PR diff**)

```bash
$ ls out/packages/endpoint/
0.18.0-dev.0  0.18.0-dev.1
$ diff -r out/packages/endpoint/0.18.0-dev.*
diff -r out/packages/endpoint/0.18.0-dev.0/elasticsearch/index_template/metrics-metadata-current.json out/packages/endpoint/0.18.0-dev.1/elasticsearch/index_template/metrics-metadata-current.json
85c85
<                                     "type": "constant_keyword"
---
>                                     "type": "keyword"
diff -r out/packages/endpoint/0.18.0-dev.0/elasticsearch/transform/metadata_current/default.json out/packages/endpoint/0.18.0-dev.1/elasticsearch/transform/metadata_current/default.json
3c3
<         "index": "metrics-endpoint.metadata-default*",
---
>         "index": "metrics-endpoint.metadata-*",
diff -r out/packages/endpoint/0.18.0-dev.0/manifest.yml out/packages/endpoint/0.18.0-dev.1/manifest.yml
5c5
< version: 0.18.0-dev.0
---
> version: 0.18.0-dev.1
```

the above showing the only difference is the contents of this PR, and the manifest version.

And the Endpoint Security integration created under a custom namespace (and an agent+endpoint spun up):

2021-01-20-124246_scrot



Here is the behavior with 0.18.0-dev.0:

2021-01-20-123311_scrot
2021-01-20-123317_scrot

2021-01-20-123358_scrot

And upon adding 0.18.0-dev.1 to the EPR, and kibana picking it up:

Kibana logs:

```
server    log   [12:38:05.561] [info][fleet][plugins] Found previous transform references:
 [{"id":"endpoint.metadata_current-default-0.18.0-dev.0","type":"transform"}]
server    log   [12:38:05.562] [info][fleet][plugins] Deleting currently installed transform ids endpoint.metadata_current-default-0.18.0-dev.0
server    log   [12:38:05.638] [info][fleet][plugins] Deleted: endpoint.metadata_current-default-0.18.0-dev.0
```

elasticsearch logs:

```
INFO [MetadataIndexTemplateService] updating index template [metrics-metadata-current] for index patterns [metrics-endpoint.metadata_current_*]
INFO [MetadataIndexTemplateService] updating index template [.logs-endpoint.diagnostic.collection] for index patterns [.logs-endpoint.diagnostic.collection-*]
INFO [MetadataIndexTemplateService] updating index template [logs-endpoint.events.library] for index patterns [logs-endpoint.events.library-*]
INFO [MetadataIndexTemplateService] updating index template [logs-endpoint.events.network] for index patterns [logs-endpoint.events.network-*]
INFO [MetadataIndexTemplateService] updating index template [metrics-endpoint.policy] for index patterns [metrics-endpoint.policy-*]
INFO [MetadataIndexTemplateService] updating index template [logs-endpoint.events.process] for index patterns [logs-endpoint.events.process-*]
INFO [MetadataIndexTemplateService] updating index template [logs-endpoint.events.registry] for index patterns [logs-endpoint.events.registry-*]
INFO [MetadataIndexTemplateService] updating index template [metrics-endpoint.metrics] for index patterns [metrics-endpoint.metrics-*]
INFO [MetadataIndexTemplateService] updating index template [logs-endpoint.events.file] for index patterns [logs-endpoint.events.file-*]
INFO [MetadataIndexTemplateService] updating index template [metrics-endpoint.metadata] for index patterns [metrics-endpoint.metadata-*]
INFO [MetadataIndexTemplateService] updating index template [logs-endpoint.alerts] for index patterns [logs-endpoint.alerts-*]
INFO [MetadataIndexTemplateService] updating index template [logs-endpoint.events.security] for index patterns [logs-endpoint.events.security-*]
INFO [IndexScopedSettings]  [.ds-metrics-endpoint.metadata-myspace-2021.01.20-000001] updating [index.default_pipeline] from [metrics-endpoint.metadata-0.18.0-dev.0] to [metrics-endpoint.metadata-0.18.0-dev.1]
INFO [IndexScopedSettings]  [.ds-metrics-endpoint.metadata-myspace-2021.01.20-000001] updating [index.default_pipeline] from [metrics-endpoint.metadata-0.18.0-dev.0] to [metrics-endpoint.metadata-0.18.0-dev.1]
INFO [IndexScopedSettings]  [.ds-metrics-endpoint.metrics-myspace-2021.01.20-000001] updating [index.default_pipeline] from [metrics-endpoint.metrics-0.18.0-dev.0] to [metrics-endpoint.metrics-0.18.0-dev.1]
INFO [IndexScopedSettings]  [.ds-metrics-endpoint.metrics-myspace-2021.01.20-000001] updating [index.default_pipeline] from [metrics-endpoint.metrics-0.18.0-dev.0] to [metrics-endpoint.metrics-0.18.0-dev.1]
INFO [IndexScopedSettings]  [.ds-metrics-endpoint.policy-myspace-2021.01.20-000001] updating [index.default_pipeline] from [metrics-endpoint.policy-0.18.0-dev.0] to [metrics-endpoint.policy-0.18.0-dev.1]
INFO [IndexScopedSettings]  [.ds-metrics-endpoint.policy-myspace-2021.01.20-000001] updating [index.default_pipeline] from [metrics-endpoint.policy-0.18.0-dev.0] to [metrics-endpoint.policy-0.18.0-dev.1]
INFO [TransformIndexer] [endpoint.metadata_current-default-0.18.0-dev.0] transform has stopped.
INFO [MetadataDeleteIndexService] [metrics-endpoint.metadata_current_default/vL-OHfA5SUaqKJV-6VT3MA] deleting index
INFO [MetadataCreateIndexService] [metrics-endpoint.metadata_current_default] creating index, cause [api], templates [metrics-metadata-current], shards [1]/[1]
INFO [AllocationService] updating number_of_replicas to [0] for indices [metrics-endpoint.metadata_current_default]
INFO [TransformTask] [endpoint.metadata_current-default-0.18.0-dev.1] updating state for transform to [{"task_state":"started","indexer_state":"stopped","checkpoint":0,"should_stop_at_checkpoint":false}].
INFO [TransformPersistentTasksExecutor] [endpoint.metadata_current-default-0.18.0-dev.1] successfully completed and scheduled task in node operation
INFO [MetadataMappingService] [metrics-endpoint.metadata_current_default/gCf0T5WhQ66DU94Q1RO_ag] update_mapping [_doc]
```

(package upgrade looks good)

2021-01-20-124041_scrot
2021-01-20-124054_scrot
2021-01-20-124105_scrot



Looks like the endpoint gets picked up just fine.

---

This seems good to go. I'm going to push up the merge conflict fix and then merge this

PR #123, opened 2021-01-11
metadata transform: only grab documents newer than 90d

pzl:
Alters the Metadata transform query to [not include documents older than 90 days](https://github.com/elastic/security-team/issues/698)

Functional proof:

https://user-images.githubusercontent.com/315796/104203128-ca3eaf80-53f9-11eb-95ee-a34273386158.mp4

### DIFF
--- a/package/endpoint/elasticsearch/transform/metadata_current/default.json
+++ b/package/endpoint/elasticsearch/transform/metadata_current/default.json
@@ -1,6 +1,13 @@
 {
     "source": {
-        "index": "metrics-endpoint.metadata-default*"
+        "index": "metrics-endpoint.metadata-default*",
+        "query": {
+            "range": {
+                "@timestamp": {
+                    "gt": "now-90d/d"
+                }
+            }
+        }
     },
     "dest": {
         "index": "metrics-endpoint.metadata_current_default"


package/endpoint/elasticsearch/transform/metadata_current/default.json
```diff
@@ -1,6 +1,13 @@
 {
     "source": {
-        "index": "metrics-endpoint.metadata-default*"
+        "index": "metrics-endpoint.metadata-default*",
+        "query": {
+            "range": {
+                "@timestamp": {
```
jonathan-buttner, line 6: Just wanted to point out that the `@timestamp` field will be the time that the elastic endpoint created the document. It won't be the time that the document was ingested by Elasticsearch. If we wanted to measure time by when the document was added to Elasticsearch then we'll probably want the `event.ingested` field instead.

jonathan-buttner (approved):
LGTM, one question about `@timestamp` vs `event.ingested`

package/endpoint/elasticsearch/transform/metadata_current/default.json
pzl: :+1: good clarification.

I would wager if the distinction were to come up, lets say two competing documents. One had a later `@timestamp`, the other had a later `event.ingested`. I would think we should go with the one with the later host-creation time.

A network / packet delay could easily place older documents arriving close together. So I would think stick with `@timestamp`

pzl: also `@timestamp` is the field we're looking at during sort & compare. Using the same field is probably the bigger consideration to stick with

PR #124, opened 2021-01-11
Change endpoint metadata transform to use "latest" operator

pzl:
the ML team added a native "latest" function of transforms, designed after our use case. This migrates our transform to start using it.

This involves a destination index schema change, Where the previous schema was:

```
{
  "agent":{ "id": "XYZ" },
  "HostDetails": { <sourcedoc> }
}
```

and agent.id is available as a field inside that sourcedoc. New schema becomes:
```
{ <sourcedoc> }
```

where `agent.id` remains available at the same place, now.

Kibana PR to sync up with the schema change: https://github.com/elastic/kibana/pull/88012

### DIFF
--- a/package/endpoint/elasticsearch/index_template/metrics-metadata-current.json
+++ b/package/endpoint/elasticsearch/index_template/metrics-metadata-current.json
@@ -23,85 +23,21 @@
                 "@timestamp": {
                     "type": "date"
                 },
-                "HostDetails": {
+                "Endpoint": {
                     "properties": {
-                        "@timestamp": {
-                            "type": "date"
-                        },
-                        "Endpoint": {
+                        "policy": {
                             "properties": {
-                                "policy": {
-                                    "properties": {
-                                        "applied": {
-                                            "properties": {
-                                                "id": {
-                                                    "type": "keyword",
-                                                    "ignore_above": 1024
-                                                },
-                                                "name": {
-                                                    "type": "keyword",
-                                                    "ignore_above": 1024
-                                                },
-                                                "status": {
-                                                    "type": "keyword",
-                                                    "ignore_above": 1024
-                                                }
-                                            }
-                                        }
-                                    }
-                                },
-                                "status": {
-                                    "type": "keyword",
-                                    "ignore_above": 1024
-                                }
-                            }
-                        },
-                        "agent": {
-                            "properties": {
-                                "id": {
-                                    "type": "keyword",
-                                    "ignore_above": 1024
-                                },
-                                "name": {
-                                    "type": "keyword",
-                                    "ignore_above": 1024
-                                },
-                                "type": {
-                                    "type": "keyword",
-                                    "ignore_above": 1024
-                                },
-                                "version": {
-                                    "type": "keyword",
-                                    "ignore_above": 1024
-                                }
-                            }
-                        },
-                        "data_stream": {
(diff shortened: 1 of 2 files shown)


pzl:
So I discovered that there is an additional coordination point to change when this gets released:

In the Kibana PR (elastic/kibana#88012) I am updating the endpoint management queries to basically check (`field || HostDetails.field`) so it works before and after schema change.

But one thing I couldn't change was the esarchive data, that is mock-injected into the transform destination index.

Because the functional test servers are using the old endpoint package and old schema. So the data in the functional test there ( `x-pack/test/functional/es_archives/endpoint/metadata/destination_index/data.json` ) is tied to the live endpoint package mapping for the transform dest index.

Meaning: when we deploy this, we will need to change this functional test data to match the deployed schema change. Or the tests in kibana CI will start failing.

kevinlog:
@pzl 

> Meaning: when we deploy this, we will need to change this functional test data to match the deployed schema change. Or the tests in kibana CI will start failing.

Something we've done in the past (not ideal) is to temporarily disable the tests right before deploying the new package.  Then, after the package is release, reenable the tests, fix, and put in a PR.

@jonathan-buttner let me know if we've come up with a better plan since

jonathan-buttner:
> @pzl
> 
> > Meaning: when we deploy this, we will need to change this functional test data to match the deployed schema change. Or the tests in kibana CI will start failing.
> 
> Something we've done in the past (not ideal) is to temporarily disable the tests right before deploying the new package. Then, after the package is release, reenable the tests, fix, and put in a PR.
> 
> @jonathan-buttner let me know if we've come up with a better plan since

hmm is this going to be a breaking change? What I mean by that is if you released these package changes _without_ your kibana changes would it break something for a developer that is running off of master?

If so, could we merge your kibana PR first (assuming it works with both the old package and the new package)?

Deploying endpoint packages shouldn't affect the kibana CI. The reason is that kibana CI runs a docker container pinned to a specific docker image which represents a specific git hash of the package-storage repo. The hash is here: https://github.com/elastic/kibana/blob/master/x-pack/test/fleet_api_integration/config.ts#L15

If you want to run the integration tests locally using the docker container this readme should help: https://github.com/elastic/kibana/blob/master/x-pack/plugins/security_solution/README.md

If you'd like to update the endpoint package used in CI so you can validate that your kibana changes work, we'll need to release a dev version of the package to `snapshot` and then you can pin the hash I mentioned above to that specific version. The snapshot branch is used by kibana's source built from master. So release a package to snapshot _could_ affect other developers.

pzl:
Yes, it is breaking.

Kibana change needs to go in first, so this PR is awaiting that (and that kibana PR _is_ intended to work with both old & new package).

jonathan-buttner:
> Yes, it is breaking.
> 
> Kibana change needs to go in first, so this PR is awaiting that (and that kibana PR _is_ intended to work with both old & new package).

Sounds good. We'll want to bump the version here too then: https://github.com/elastic/endpoint-package/blob/master/package/endpoint/manifest.yml#L19 so the new package only gets applied to 7.12 stacks.

pzl:
The partner Kibana PR to this has been merged to master and 7.x

So this is ready for 1) review  and 2) aggressive testing

jonathan-buttner (approved):
I compared it with the old mapping and it looks good 👍

PR #128, opened 2021-01-13
Enabling linting and updating elastic-package

jonathan-buttner:
This PR enables linting using the `elastic-package` tool. It was temporarily disabled because the `unsigned_long` type added in this PR was failing the linter: https://github.com/elastic/endpoint-package/pull/127/files#diff-f22724a63347919299942eaa577a6c6081212bd8a91c18cdc5a391a63e3798e6R186

We had to wait for the `elastic-package` tool to be updated before we could enabled it again. I also added a target to the Makefile to make updating the tool easier in the future.

The vendor files are changing because we pin the `elastic-package` tool to a specific version in this repo.


pzl (approved):
lgtm

PR #134, opened 2021-02-16
add transform retention policy (delete >90d)

pzl:
This deletes endpoint metadata documents when they are over 90d.

This uses the transform-native feature that was added: "retention policy".

Installing a transform with that definition went fine. Set one to 89d old, waiting for a day rollover to see if it gets deleted

2021-02-16-090803_scrot
2021-02-16-090825_scrot

### DIFF
--- a/package/endpoint/elasticsearch/transform/metadata_current/default.json
+++ b/package/endpoint/elasticsearch/transform/metadata_current/default.json
@@ -18,6 +18,12 @@
         ],
         "sort": "@timestamp"
     },
+    "retention_policy": {
+        "time": {
+            "field": "event.ingested",
+            "max_age": "90d"
+        }
+    },
     "description": "collapse and update the latest document for each host",
     "frequency": "1m",
     "sync": {


jonathan-buttner (approved):

pzl:
It's rollover day.

2021-02-17-084331_scrot
2021-02-17-084344_scrot
2021-02-17-084359_scrot


Curiously, `documents_deleted` is still listed as `0` in stats, but also, `delete_time_in_ms` is 4.. so maybe it still ran and didn't update the count.

I'm not sure if these ES logs are relevant, or something else:

```
INFO [MlDailyMaintenanceService] triggering scheduled [ML] maintenance tasks
INFO [TransportDeleteExpiredDataAction] Deleting expired data
INFO [UnusedStatsRemover] Successfully deleted [0] unused stats documents
INFO [TransportDeleteExpiredDataAction] Completed deletion of expired ML data
INFO [MlDailyMaintenanceService] Successfully completed [ML] maintenance task: triggerDeleteExpiredDataTask
```

_Regardless_ of the reporting, the functionality happened, and the document is not present in the transform destination.

kevinlog:
this is awesome, thanks @pzl 

I'm thinking we should target this for `7.13` so that we can adjust our Endpoint list query to search for Agents only as far back as `89d`.  This would keep us from getting out of sync in the UI as shown in your Endpoint list screenshot above.  I hope that this would be a rare occurrence, but with large deployments my guess is that old Agents will be laying around for awhile.  

Alternatively, we could see if they're planning a retention on the Agent/Fleet side.

One more question - this just ensures that 90d old doesn't appear in the transform destination, correct?

In terms of merging, I'm OK with it, I just think we should hold off on releasing it until we address the above.

pzl:
one part is already merged that would go out for `7.12`: the transform will not *pick up* documents from the source docs that are _already_ >90d old. 

But if something is added to the destination index when it is fresh, goes offline, 90d pass, it won't be removed. This PR does that half.


So we still have the possibility of the erroneous enrolling message, if there exist old endpoint docs.. and the dest gets wiped some other way (package upgrade, manual, idk).

If you want to hold this half to not go into `7.12`, i can just hold off on merging, and we can cut/deploy the 7.12 now, and merge to master after. But half the feature is already slotted for `.12`

PR #137, opened 2021-03-25
Add DLL load_index

gabriellandau:
Relates to https://github.com/elastic/endpoint-dev/issues/8659

It can be useful to know when certain system DLLs (ntdll, kernel32) have been loaded a second time into any given process.  This can indicate malicious behavior.  These DLLs are normally loaded into each Windows process once at startup, so a "process creation" event generally implies one load of each of these DLLs, and is thus uninteresting and noisy.  Discussions with @joe-desimone led to a decision to only report on the second load of those two DLLs, to reduce noise and data volume.  I've already implemented this in the Endpoint here: https://github.com/elastic/endpoint-dev/pull/8661

This will eventually lead to rules where we say "I want to know about the load of kernel32", but only the second.  We need some way to clearly indicate to the user that this is in fact the second load of kernel32.  `load_index` is that indicator.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,2 +1,4 @@
 generated/** linguist-generated=true
-
+package/** linguist-generated=true
+package/endpoint/docs/README.md linguist-generated=true
+schemas/**
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+generated/* linguist-generated=true
+
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,2 +1,2 @@
-generated/* linguist-generated=true
+generated/** linguist-generated=true
 
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,4 +1,4 @@
 generated/** linguist-generated=true
 package/** linguist-generated=true
 package/endpoint/docs/README.md linguist-generated=true
-schemas/**
+schemas/** linguist-generated=true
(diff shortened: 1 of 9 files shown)


gabriellandau:
Running `make` against `master` yields the following changed files:
```
user@ubuntu:~/git/endpoint-package$ git status
On branch master
Your branch is up to date with 'origin/master'.

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git checkout -- <file>..." to discard changes in working directory)

        modified:   generated/alerts/ecs/subset/malware_event/exceptionable.json
        modified:   generated/alerts/ecs/subset/ransomware_event/exceptionable.json

no changes added to commit (use "git add" and/or "git commit -a")
```

Running `make` against this branch yields the same two changed files.  Should my changes to `custom_schemas/custom_dll.yml` affect any other files in `generated/`?

.gitattributes
```diff
@@ -0,0 +1,4 @@
+generated/** linguist-generated=true
+package/** linguist-generated=true
+package/endpoint/docs/README.md linguist-generated=true
+schemas/** linguist-generated=true
```
gabriellandau, line 4: This tells GitHub that these files are generated, which will reduce redundancy in the PR and hopefully let us focus on the changes that matter.  GitHub docs: https://docs.github.com/en/github/administering-a-repository/customizing-how-changed-files-appear-on-github

marshallmain (approved):

.gitattributes
jonathan-buttner: Thanks for adding these @gabriellandau. I think we might want to leave the one for `package/**` so that we can ensure that the generated files look correct. It's possible we still have some bugs in the generation code 😄

gabriellandau: Roger.  See https://github.com/elastic/endpoint-package/pull/140

PR #139, opened 2021-04-01
Add defense evasion list to process creation events

gabriellandau:
The description of the new field should explain this.  If it doesn't, I should update that instead of explaining it here.

### DIFF
--- a/custom_schemas/custom_process.yml
+++ b/custom_schemas/custom_process.yml
@@ -186,7 +186,7 @@
         
     - name: Ext.defense_evasions
       level: custom
-      type: array
+      type: keyword
       short: List of defense evasions found in this process.
       description: >
         List of defense evasions found in this process.  
--- a/custom_schemas/custom_process.yml
+++ b/custom_schemas/custom_process.yml
@@ -183,3 +183,12 @@
       short: The real pid of the process if ppid spoofing is happening.
       description: >
         For process.parent this will be the ppid of the process that actually spawned the current process.
+        
+    - name: Ext.defense_evasions
+      level: custom
+      type: array
+      short: List of defense evasions found in this process.
+      description: >
+        List of defense evasions found in this process.  
+        These defense evasions can make it harder to inspect a process and/or cause abnormal OS behavior.
+        Examples tools that can cause defense evasions include Process Doppelgänging and Process Herpaderping.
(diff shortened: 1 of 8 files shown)


custom_schemas/custom_process.yml
```diff
@@ -183,3 +183,12 @@
       short: The real pid of the process if ppid spoofing is happening.
       description: >
         For process.parent this will be the ppid of the process that actually spawned the current process.
+        
+    - name: Ext.defense_evasions
+      level: custom
+      type: array
```
jonathan-buttner, line 189: The `type` here needs to be an Elasticsearch type: https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping-types.html

Is this an array of strings? `keyword` will work if so, because everything in Elasticsearch is an array essentially.

gabriellandau: > Is this an array of strings? keyword will work if so, because everything in Elasticsearch is an array essentially.

Yep.  Thanks.  Fixed.  https://github.com/elastic/endpoint-package/pull/139/commits/8a61318c91ad14a8b39ff0d514d12c3c30d3c501

Maybe `make` should fail in such cases?

jonathan-buttner: Yeah that's a good idea. I'll create an issue so we can track it.

custom_subsets/elastic_endpoint/process/process.yaml
```diff
@@ -150,6 +150,7 @@ fields:
               subject_name: {}
               trusted: {}
               valid: {}
+          defense_evasions: {}
```
jonathan-buttner, line 153: Just to make sure, will the endpoint only send the `defense_evasions` for process events? Would this field show up in any alert documents?

If so we might need to add it to a couple places in the alerts directory like here:
https://github.com/elastic/endpoint-package/blob/8a61318c91ad14a8b39ff0d514d12c3c30d3c501/custom_subsets/elastic_endpoint/alerts/malware_event.yaml#L354

gabriellandau: > Would this field show up in any alert documents?

No plans for that currently.  For 7.13 at least, it'll only be in process creation events.

jonathan-buttner: ok cool 👍

jonathan-buttner (approved):

gabriellandau:
Thanks @jonathan-buttner

PR #141, opened 2021-04-07
ISSUE-8721: add event.risk_score to malware event subset

brian-mckinney:
Fixing endpoint schema violations.  This fixes event.risk_score, which is part of ECS proper, but was not added to the malware event custom subset, so validation was faling.


custom_schemas/custom_event.yml
```diff
@@ -18,6 +18,11 @@
     metric and state events.
   type: group
   fields:
+    - name: risk_score
```
jonathan-buttner, line 21: Looks like `risk_score` is actually included in the core ECS implementation: https://github.com/elastic/ecs/blob/master/schemas/event.yml#L678

So I think we can skip having to define it here as a _custom_ event field.

If you remove the changes in this file and rerun `make` it should still include the changes below except that it'll be `type: float` instead of `type: long`.

Is it an issue if we map this as a `float` instead of a `long`?

brian-mckinney: I must have added this change by accident when I was learning the file structure.  It shouldn't be here.  I removed it.

brian-mckinney:
FYI, I also followed the instructions for signing the contributor agreement

package/endpoint/data_stream/alerts/fields/fields.yml
```diff
@@ -1752,6 +1752,10 @@
 
         Event transports such as Syslog or the Windows Event Log typically mention the source of an event. It can be the name of the software that generated the event (e.g. Sysmon, httpd), or of a subsystem of the operating system (kernel, Microsoft-Windows-Security-Auditing).'
       example: kernel
+    - name: risk_score
+      level: custom
+      type: long
```
jonathan-buttner, line 1757: hmm I think should switch to a `float` now 🤔 , if you run `make` does it show any additional changes?

brian-mckinney:
I'm gonna close this PR and submit the changes to another PR to avoid the CLA checker issue.

PR #144, opened 2021-04-12
Schema for Memory Protection Alerts

gabriellandau:
Actually add schema for Memory Protection alerts, and add a new `memory_region` message that can be shared with MemoryScanner.


custom_schemas/custom_memory_region.yml
```diff
@@ -0,0 +1,125 @@
...
+  group: 2
+  short: Fields describing a memory region.
+  description: >
+    Fields describing a memory region.
+  reusable:
+    order: 1
+    top_level: false
+    expected:
+      - at: process.thread.Ext
+        as: start_address_details
```
gogochan, line 13: This is awesome. for memory_scan, I can add another value after this PR gets merged.
```
- at: process.Ext.malware_signature
  as: memory_region
```

custom_schemas/custom_process.yml
```diff
@@ -108,12 +149,36 @@
...
+      level: custom
+      type: long
+      description: >
+        When a thread is created, this is the raw numerical value of its parameter.
+
+    - name: thread.Ext.parameter_bytes_compressed
+      level: custom
+      type: keyword
+      description: >
+        Up to 512KB of raw data from the thread parameter, if it is a valid pointer. This is compressed with zlib. To reduce data volume, this is de-duplicated on the endpoint, and may be missing from many alerts if the same data would be sent multiple times.
```
jonathan-buttner, line 163: Will the endpoint send this value at least once for every process on the host?

custom_schemas/custom_process.yml
```diff
@@ -108,12 +149,36 @@
...
 
     - name: Ext.authentication_id
       level: custom
       type: keyword
       description: >
         Process authentication ID
+        
+    - name: Ext.dll
+      level: custom
+      type: nested
```
jonathan-buttner, line 179: Could we use an `object` here or do we anticipate needing to query fields on the `Ext.dll` together? 

Docs: https://www.elastic.co/guide/en/elasticsearch/reference/current/nested.html

On the `Target` field set we have it mapped as an object I think.

custom_subsets/elastic_endpoint/alerts/memory_protection_event.yaml
```diff
@@ -0,0 +1,470 @@
...
+      name: {}
+      Ext:
+        fields:
+          real:
+            fields:
+              id:
+                exceptionable: true
+              name: {}
+  process:
+    fields: &process-fields
```
jonathan-buttner, line 197: 💯

jonathan-buttner (commented):
Thanks for putting this together Gabe!

custom_schemas/custom_dll.yml
```diff
@@ -23,7 +24,7 @@
 
     - name: Ext.mapped_address
       level: custom
-      type: keyword
+      type: long
```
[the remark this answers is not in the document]
gabriellandau: I could imagine wanting to do math on these fields.  If it breaks kibana, that's a deal-breaker though.  

I see this "unsigned long" type.  Would it be okay to use this?  https://www.elastic.co/guide/en/elasticsearch/reference/current/unsigned-long.html

custom_schemas/custom_memory_region.yml
```diff
@@ -0,0 +1,125 @@
...
+    - name: region_protection
+      level: custom
+      type: keyword
+      example: "RWX"
+      description: >
+        Memory protection of the memory region. Example values include "RWX" and "R-X".
+
+    - name: region_size
+      level: custom
+      type: long
```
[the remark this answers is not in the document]
gabriellandau: Okay thanks.  I'm going to change all these to `unsigned_long`

custom_schemas/custom_process.yml
gabriellandau: Not necessarily.  It will send it once per unique injection fingerprint (`unique_key_v1`).  It's unique to the memory injected, not actor or target process/pid/etc.

custom_schemas/custom_memory_region.yml
```diff
@@ -0,0 +1,125 @@
...
+  reusable:
+    order: 1
+    top_level: false
+    expected:
+      - at: process.thread.Ext
+        as: start_address_details
+  fields:
+    - name: allocation_base
+      level: custom
+      type: long
```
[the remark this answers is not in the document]
gabriellandau: Switched to `unsigned_long`

custom_schemas/custom_memory_region.yml
```diff
@@ -0,0 +1,125 @@
...
+    - name: allocation_type
+      level: custom
+      type: keyword
+      example: "PRIVATE"
+      description: >
+        The memory allocation type. Example values include "IMAGE", "MAPPED", and "PRIVATE".
+        
+    - name: bytes_address
+      level: custom
+      type: long
```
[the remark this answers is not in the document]
gabriellandau: Switched to `unsigned_long`

custom_schemas/custom_memory_region.yml
```diff
@@ -0,0 +1,125 @@
...
+      example: 2431737462784
+      description: >
+        The address where bytes_compressed begins.
+        
+    - name: bytes_allocation_offset
+      level: custom
+      type: long
```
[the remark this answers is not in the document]
gabriellandau: Switched to `unsigned_long`

custom_schemas/custom_process.yml
```diff
@@ -84,18 +91,52 @@
 
     - name: thread.Ext.start_address
       level: custom
-      type: keyword
-      example: 5442508
+      type: long
```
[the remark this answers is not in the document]
gabriellandau: Switched to `unsigned_long`

custom_schemas/custom_process.yml
```diff
@@ -12,6 +12,7 @@
   reusable:
     top_level: true
     expected:
+      - { at: process, as: parent }
```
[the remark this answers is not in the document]
gabriellandau: For alerts (not process events), Endpoint is sending the exact same data for `process`, and `process.parent` EXCEPT that `process.parent.parent` is not populated (no recursion).

custom_schemas/custom_process.yml
gabriellandau: I don't have a current use for nested queries here.  This was my unsuccessful attempt to get the Endpoint schema validator to recognize `process.Ext.dll`:  https://gist.github.com/gabriellandau/5e0e3b9bdfbdf336666e84f8735bcb5f#file-gistfile1-txt-L2474-L2477

I gave up and added it as a [schema violation](https://github.com/elastic/endpoint-dev/blob/6b8cc335fb3b945c8f85bfc9e5d1c624d64f28a4/ElasticCommonSchema/schema_violations#L57-L67) for now, since Memory Protection alerts are not customer-facing, and getting this in will enable @gogochan to make progress.

While I've got your attention, can you [take a look](https://gist.github.com/gabriellandau/5e0e3b9bdfbdf336666e84f8735bcb5f#file-gistfile1-txt-L2851-L2924) and tell me if you see anything wrong?

I'm happy to take this out since it's not helping.

custom_schemas/custom_dll.yml
```diff
@@ -23,13 +24,13 @@
 
     - name: Ext.mapped_address
       level: custom
-      type: keyword
+      type: unsigned_long
```
jonathan-buttner, line 27: Another thing I thought about related to these change. Since this is a type change, I believe on upgrade the old data stream will get rolled over because of the mapping conflict. That _should_ work but we'll want to test that upgrading to this package works ok.

custom_schemas/custom_process.yml
jonathan-buttner: Gotcha, yeah I just wanted to make sure that we weren't going to be sending hundreds of KB of data on each process event.

jonathan-buttner (approved):
This looks good to me. I tested the package can install on a fresh version of Kibana running locally. We'll want to make sure that upgrading to this package works correctly because of the `type` changes in a couple places.

image

custom_schemas/custom_process.yml
gabriellandau: > on each process event.

So even though this is in `custom_process.yml`, it's for the `process` field within Memory Protection alerts (currently only shellcode injection).  `process.thread` won't be populated within process events.

custom_schemas/custom_dll.yml
gabriellandau: All of the `unsigned_long` fields here are not new to 7.13, but unused in 7.12 (only in diagnostics which had no mapping).  Does that mean we're okay?

- `dll.Ext.mapped_address`
- `dll.Ext.mapped_size`
- `process.thread.Ext.start_address`

custom_schemas/custom_process.yml
gabriellandau: If we were to use `object` now then later change it to `nested`, would Kibana migrate it seamlessly?

gabriellandau: I changed this to `object`.  Per [discussions](https://elastic.slack.com/archives/C8L4TMJGJ/p1618413271200900) in [Slack](https://elastic.slack.com/archives/C8L4TMJGJ/p1618413576202300), we can later switch this to `nested` and ES will automatically roll over the data stream.  `nested` queries may return FPs on older data, but it seems that's about the extent of the risk.  

Let's stick with `object` for now.

custom_schemas/custom_dll.yml
marshallmain: Even without data being streamed into the indices, the mapping conflict means aggregations won't work on this field if a query includes indices that have it mapped both ways. I think that's fine, since it's not populated in 7.12 so we shouldn't be breaking any existing queries, and once the old indices roll off then it won't be an issue at all. So I think it's good to update this to `unsigned_long`, just noting what the short term tradeoff could be with aggregations when changing field types in general.

custom_schemas/custom_memory_protection.yml
```diff
@@ -0,0 +1,39 @@
...
+    
+    - name: self_injection
+      level: custom
+      type: boolean
+      description: Is this alert about a process injecting into itself?
+
+    - name: parent_to_child
+      level: custom
+      type: boolean
+      description: Is this process injecting into its child?
```
marshallmain, line 29: Is it too late to consider merging these 2 fields into one `keyword` field that could be an enum of string values? e.g. `injection_target: "self"`, `injection_target: "child"`, or 'injection_target: "other"`. Having these in a single field would make it easier for analysts to write queries that make use of the injection target.

custom_schemas/custom_process.yml
gabriellandau: After doing more reading, I learned that with `object`s loss of ability to correlate fields within a sub-message isn't just limited to the index/search.  It's also lost in the document if you pull it out: https://www.elastic.co/guide/en/elasticsearch/reference/current/nested.html#nested-arrays-flattening-objects

This means we may not be able to cleanly correlate fields in the offline python scripts that we use to triage alerts.

Example data:
```
                    "dll": [
                        {
                            "Ext": {
                                "mapped_address": 6225920,
                                "mapped_size": 122880
                            },
                            "code_signature": [
                                {
                                    "exists": false,
                                    "status": "noSignature",
                                    "subject_name": "",
                                    "trusted": false
                                }
                            ],
                            "hash": {
                                "md5": "b2d4e40a3ab592e7c0e5ed353ab85836",
                                "sha1": "7a7fbc6d0f4bdc225bfda2710611135aa79d9a6f",
                                "sha256": "c8fd1c0a2db5706798f0b2c9e64afb37e09465d7198d2ad8e47b5b288f526e64"
                            },
                            "name": "test_shellcode_thread.exe",
                            "path": "C:\\git\\endpoint-dev\\LfsFileDeps\\Tools\\test_shellcode_thread.exe"
                        },
```

I'm having second thoughts about `object`.  `nested` may be the way to go.

custom_schemas/custom_memory_region.yml
```diff
@@ -0,0 +1,125 @@
...
+      example: 4096
+      description: >
+        Original memory protection requested when the memory was allocated. Example values include "RWX" and "R-X". 
+        
+    - name: allocation_type
+      level: custom
+      type: keyword
+      example: "PRIVATE"
+      description: >
+        The memory allocation type. Example values include "IMAGE", "MAPPED", and "PRIVATE".
```
marshallmain, line 41: these 4 look like good candidates to group in an object `allocation.*`

custom_schemas/custom_memory_region.yml
```diff
@@ -0,0 +1,125 @@
...
+      example: "eJzzSM3JyVcIzy/KSVEEABxJBD4="
+      description: >
+        Up to 4MB of raw data from the memory allocation. This is compressed with zlib. To reduce data volume, this is de-duplicated on the endpoint, and may be missing from many alerts if the same data would be sent multiple times.
+        
+    - name: bytes_compressed_present
+      level: custom
+      type: boolean
+      example: false
+      description: >
+        Whether bytes_compressed is present in this event.
```
marshallmain, line 69: good candidates to group in `bytes.*`

custom_schemas/custom_process.yml
```diff
@@ -84,18 +91,52 @@
...
+      description: >
+        The bytes at the thread start address, disassembled into human-readable assembly code.
+        
+    - name: thread.Ext.start_address_bytes_disasm_hash
+      level: custom
+      type: keyword
+      example: "aacb1c801f9030f799e2f7350f053ebb760d42cbe81cd65021063c1c4d1a9c9c"
+      description: >
+        The bytes at the thread start address, with immediate values capped to 0x100, disassembled into human-readable assembly code, then hashed.
+
```
marshallmain, line 126: Might be too late since `start_address` is a pre-existing field, but grouping these in a `start_address` object would be nice

custom_schemas/custom_memory_region.yml
```diff
@@ -0,0 +1,125 @@
...
+      example: 0
+      description: >
+        Offset of bytes_address the memory allocation. Equal to bytes_address - allocation_base.
+        
+    - name: bytes_compressed
+      level: custom
+      type: keyword
+      example: "eJzzSM3JyVcIzy/KSVEEABxJBD4="
+      description: >
+        Up to 4MB of raw data from the memory allocation. This is compressed with zlib. To reduce data volume, this is de-duplicated on the endpoint, and may be missing from many alerts if the same data would be sent multiple times.
```
marshallmain, line 62: We may want to add `index: false` to this field, as indexing compressed data is probably unnecessary.

custom_schemas/custom_process.yml
```diff
@@ -108,12 +149,36 @@
...
+    - name: thread.Ext.parameter
+      level: custom
+      type: unsigned_long
+      description: >
+        When a thread is created, this is the raw numerical value of its parameter.
+
+    - name: thread.Ext.parameter_bytes_compressed
+      level: custom
+      type: keyword
+      description: >
```
marshallmain, line 162: consider adding `index: false` so we don't index compressed data

marshallmain (commented):
Looks find overall, a couple comments on potential improvements. I'm not sure how much effort it would take to group some of the fields together in objects, if these field names have already been decided elsewhere then it's not a big issue.

custom_schemas/custom_process.yml
marshallmain: It depends on how the document is pulled out: the `_source` field preserves the original JSON document that was provided to ES to index, and will therefore preserve correlation between fields within a sub-message if you retrieve the document using `_source`. Other methods of retrieving fields within a document such as the [fields parameter](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-fields.html) or [doc value fields](https://www.elastic.co/guide/en/elasticsearch/reference/master/search-fields.html#docvalue-fields) would *not* preserve the correlation. However, `_source` is the typical easy way I would expect users to get a full document for use in offline scripts so `object` should work here.

custom_schemas/custom_memory_protection.yml
gabriellandau: It's too late for 7.13.  Endpoint feature freeze is tomorrow.  This is still diagnostic-only, so changing it before 7.14 is an option.

gabriellandau: We do have a corpus of exceptionlist entries based off this schema: https://github.com/elastic/global-security-exceptionlist/blob/629cc63b2558ec2d57f7fa1b3d212cfb19c78ccc/entries/windows/processInjection.yml#L423

They would have to be updated if we changed this.

custom_schemas/custom_memory_region.yml
gabriellandau: Thanks.  Possibly for 7.14, but this won't make the 7.13 train.

pzl (approved):
:+1:  lgtm

we should do upgrade tests after merging as @jonathan-buttner called out, but otherwise :ok_hand:

gabriellandau:
@jonathan-buttner @pzl ready for upgrade testing.

custom_schemas/custom_memory_protection.yml
jdu2600: Do we want to make this change?

gabriellandau: @jdu2600 I didn't back in April when this was merged.  Do you want it?

jdu2600: When I was writing exception list rules it definitely felt to me like it would be significantly cleaner to have this in a single three-valued field like @marshallmain suggested. 

But is that enough justification for the work required to make this change and migrate the exception list rules (and test samples)...

It might be useful to get a 'customer' perspective from our I&A team? @devonakerr @Samirbous

devonakerr:
I didn't see much discussion so let me cc @brokensound77 and @dstepanic17 to share any thoughts they might have.

PR #145, opened 2021-04-29
Add Isolation fields to metadata & policy schemas

pzl:
Adds:

- `Endpoint.configuration` to show what non-policy settings are set to be. 
- `Endpoint.state` to reflect what the current status of those same fields are

These are intended to have the same fields in each, to reflect any drift between a configured state and what's actually happening. 

This change adds `isolation` as the first setting, as a boolean. The falsy case `null / false` is backwards compatible, as that would reflect the correct configuration state and current state for endpoints not supporting isolation yet.


custom_schemas/custom_host.yml
```diff
@@ -0,0 +1,20 @@
...
+    A host is defined as a general computing instance.
+
+    ECS host.* fields should be populated with details about the host on which the event happened, or from which the measurement was taken. Host types include hardware, virtual machines, Docker containers, and Kubernetes nodes.
+  type: group
+  fields:
+    - name: Ext
+      level: custom
+      type: object
+      description: Object for all custom defined fields to live in.
+    - name: Ext.isolated
```
jonathan-buttner, line 16: I was reading through the ECS docs about host and just wanted to double check that `isolated` made sense under this namespace.

I haven't checked with the ECS team and this is an extension field anyway but the host fields seem to be related to `hardware, virtual machines, Docker containers, and Kubernetes nodes.`

https://www.elastic.co/guide/en/ecs/current/ecs-host.html

Do you think it'd make more sense to have the `isolation` field under the `Endpoint` section since it was an action that the endpoint did on the host?

pzl: Valid point. I was waffling back and forth on this too, and I thought `host` was more reasonable because it is _not_ a state of being of the `Endpoint` itself. It is actually changing the network state of the entire host machine, inclusive of all beats, processes, etc.

Endpoint did initiate the isolation, but the machine itself is the thing affected

jonathan-buttner (dismissed):

ferullo:
Endpoint currently doesn't actually report the configuration it is in. It reports if it applied the configuration successfully or unsuccessfully, but not what the actual situation is. That information is derivable by looking at the policy status and checking Kibana to see what the policy settings are. There is no "one stop shop" to see an actual protection/setting at the moment. It seems beneficial to add that long term for all settings, not just isolation.

With that in mind, I also think it makes sense to go inside the `Endpoint` section (`Endpoint.policy.status.isolated=true|false`?) since reporting other settings like `Endpoint.policy.status.malware=detect|prevent|off|whatever` feel analogous to reporting the isolation status.

pzl:
Good to know what the value coming back means. And certainly true that we should consider the expansion of other statuses inside Endpoint.policy.

One caveat in this case, is that isolation in Elastic is not tied to policy. It is not sent down in policy configuration, is not stored in the policy configuration in Kibana/ES. But, understood that this is a "configuration" thing, similar to policy, just sent out-of-band from policy itself. And the reporting is whether the configuration was successful

`Endpoint.isolated` sounds ok to me. Other things that are tied to policy will end up under `Endpoint.policy`

jonathan-buttner (approved):

PR #149, opened 2021-05-03
ISSUE-8721: schema violations

brian-mckinney:
https://github.com/elastic/endpoint-dev/issues/8721

Fixes the majority of violations found in:
 https://github.com/elastic/endpoint-dev/blob/master/ElasticCommonSchema/schema_violations


custom_schemas/custom_endpoint.yml
```diff
@@ -63,109 +63,211 @@
...
+      type: keyword
+      description: the version of this applied policy
+
     - name: policy.applied.response
       level: custom
       type: object
       enabled: false
       description: the response of actions that failed in the applied policy
 
-    - name: policy.applied.configurations
```
jonathan-buttner, line 72: So `policy.applied.configurations` is not a valid object? Or does it exist and `policy.applied.response.configurations` exists?

custom_schemas/custom_file.yml
```diff
@@ -27,6 +27,55 @@
         - name: text
           type: text
 
+    - name: code_signature
+      level: custom
+      type: nested
```
jonathan-buttner, line 32: ECS core already includes `file.code_signature` but it is mapped an `object` field. If we want this to be `nested` field like in other places we'll need this to be under the `Ext` object below.

Example: https://github.com/elastic/endpoint-package/blob/master/custom_schemas/custom_dll.yml#L43
ECS Core: https://github.com/elastic/ecs/blob/master/schemas/code_signature.yml

package/endpoint/data_stream/alerts/fields/fields.yml
```diff
@@ -2818,6 +2828,44 @@
         Attributes names will vary by platform. Here''s a non-exhaustive list of values that are expected in this field: archive, compressed, directory, encrypted, execute, hidden, read, readonly, system, write.'
       example: '["readonly", "system"]'
       default_field: false
+    - name: code_signature.exists
```
jonathan-buttner, line 2831: The generated files aren't taking `code_signature` as a `nested` type here because of the collision I think. If it had it would look like process:

https://github.com/elastic/endpoint-package/blob/6c4b1cb25d2588d636fe89d3edefece0bfa13775/package/endpoint/data_stream/process/fields/fields.yml#L515-L519

generated/alerts/ecs/subset/malware_event/exceptionable.json
```diff
@@ -1,234 +1,4 @@
 [
-  "Endpoint.policy.applied.id",
```
jonathan-buttner, line 2: @marshallmain this seems odd to me, do these removals look correct given the minor changes to the `malware_event` schema additions?

custom_schemas/custom_process.yml
```diff
@@ -181,6 +181,35 @@
       description: >
         List of modules loaded within this process
 
+    - name: Ext.dll.name
```
jonathan-buttner, line 184: @gabriellandau PR (https://github.com/elastic/endpoint-package/pull/144/files#diff-35f96496dd03f0c0e95a01f6d634b1959001c636d16bdab6ec357f9b64c4e104R18) added the dll fields as reusable at the location `process.Ext` so I _think_ it should work without the additions here.

Can you try removing `Ext.dll.*` but leave the subset file the way it is and see if it still pulls in the changes here: https://github.com/elastic/endpoint-package/pull/149/files#diff-163130a32df44e71133efb7f5f6e8d80f5837311251f92c2d78413d5560a8a35R564

jonathan-buttner (changes requested):
Good stuff! Just a few questions and changes.

jonathan-buttner:
@pzl could you take a look at the policy changes? I'm not as familiar with those structures.

pzl:
policy changes look good from an initial look, I asked @parkiino to double check, since she did the UI for it, and would know if those fields match what kibana is using now

paul-tavares:
Took a quick glance and certainly all of the policy response fields (`Endpoint.policy.applied**`) that were shifted will impact Kibana and the Policy Response view (I think/hope it might be isolated, however, to just one module (`selectors`).  @parkiino is certainly more familiar with the code there, but essentially, we're working with this Type:

https://github.com/elastic/kibana/blob/4ff9bfd113609fccc51fe314b79447f55a4d6cc6/x-pack/plugins/security_solution/common/endpoint/types/index.ts#L997-L1055 

So changes to the above type will give us a better idea of the impact, but I'm still thinking its mostly the selectors. 

I don't think our team (Onboarding and lifecycle team) uses the other fields listed in this PR. 

Questions:

1. what release is this for?
2. How will migrations be handled (we have to assume that old and new endpoint will be streaming data)?


//cc: @kevinlog

parkiino:
The types file that Paul put up confirms that the fields the UI is expecting is already matching the new changes, so it doesn't look like we'd need to make any changes on our end. A little confused as to how the UI type was already matching the schema if `response` wasn't already added to the path

pzl:
The documents we were receiving were _not_ matching the schema, likely. This is what this correction PR is for.

Looks like this is good aside from the callouts from @jonathan-buttner

jonathan-buttner:
> The documents we were receiving were _not_ matching the schema, likely. This is what this correction PR is for.
> 
> Looks like this is good aside from the callouts from @jonathan-buttner

Thanks @pzl @parkiino @paul-tavares ! Yeah the endpoint is probably sending them at the right location already but we hadn't mapped them there. A follow up question then, do we need them mapped? I'm fine with it either way, seems like we aren't searching on this data so far right?

> what release is this for?

@brian-mckinney is it ok if these go into 7.14?

> How will migrations be handled (we have to assume that old and new endpoint will be streaming data)?

Sounds like we don't need a migration. All the fields are additions except the policy changes which sounds like should have been there already and the old locations aren't being used.

custom_schemas/custom_endpoint.yml
brian-mckinney: from what I was told, `policy.applied.*` was changed to `policy.applied.response.*`

ferullo: Endpoint has always sent as `policy.applied.response.*`. Whether the schema was wrong to what we all agreed to or we agreed to `policy.applied.*` and Endpoint and Kibana then accidentally used `policy.applied.response.*` I don't know. But this doesn't change any data from what Endpoint sends, it just puts the schema in line with the actual layout since 7.9.

jonathan-buttner: Got it 👍  Thanks for the response. I think the way you have it here Brian looks good then.

custom_schemas/custom_endpoint.yml
```diff
@@ -271,6 +373,20 @@
         The counts field includes 20 buckets of how many times the endpoint's cpu usage fell into each bucket
       short: CPU histogram
 
+    - name: metrics.cpu.endpoint.histogram.values
```
jonathan-buttner, line 376: hmm I don't think we want to map `.values` and `.counts`. These should already be mapped by the field above:

`metrics.cpu.endpoint.histogram` since it is marked as `type: histogram`.

I think we might need to update the validator to ignore this field because a `histogram` field is defined to have two fields inside of it: `values` and `counts`

So by defining `metrics.cpu.endpoint.histogram` as a `histogram` Elasticsearch will know that these fields should be mapped automatically.

https://www.elastic.co/guide/en/elasticsearch/reference/master/histogram.html

custom_subsets/elastic_endpoint/alerts/malware_event.yaml
```diff
@@ -199,6 +200,17 @@ fields:
       accessed: {}
       attributes:
         exceptionable: true
+      code_signature:
```
jonathan-buttner, line 203: Just wanted to double check, is the endpoint going to send an array of `code_signatures` here? If it is we might to move this to an extension field instead because we won't be able to query the signatures individual since it isn't marked as type `nested` in core ECS.

jonathan-buttner (commented):
Thanks for addressing my earlier comments! I left a couple more, overall it's looking good.

custom_schemas/custom_endpoint.yml
brian-mckinney: I made the changes for this and added values and counts back to the ignored list.  Jenkins run was good.

custom_subsets/elastic_endpoint/alerts/malware_event.yaml
brian-mckinney: I just verified with @ferullo on this.  `file.code_signature` is a single value, `file.Ext.code_signature` is an array.  `file.Ext.code_signature` was already present in this subset file.

jonathan-buttner (approved):

brian-mckinney:
@jonathan-buttner Are there any further requirements for merging this PR?

jonathan-buttner:
> @jonathan-buttner Are there any further requirements for merging this PR?

Nope! Feel free to merge it.

PR #151, opened 2021-06-02
File Header bytes name and format change

AaronLamb1:
Changing file.Ext.header_data as an array of decimals to 
file.Ext.header_bytes as a hex string like this:

"header_bytes": "7468697320697320612073696C6C7920",

I flipped a coin on using keyword or text, feel free to point out a better type to use for this.


custom_schemas/custom_file.yml
```diff
@@ -174,9 +174,9 @@
       description: >
         Entropy calculation of file's header and footer used to check file integrity.
 
-    - name: Ext.header_data
+    - name: Ext.header_bytes
```
jonathan-buttner, line 177: Did we writing detection rules against `header_data` in a previous version of Kibana? Or was it just for debugging? Are we planning to write rules against `header_bytes`? If we wrote rules using `header_data` we might want to leave it around even if the endpoint doesn't continue to send data to that field otherwise the rules will likely have issues.

custom_schemas/custom_file.yml
```diff
@@ -174,9 +174,9 @@
...
       level: custom
-      type: text
+      type: keyword
```
jonathan-buttner, line 179: My guess is `keyword` is the way to go here. Here are some Elasticsearch docs about the difference between the two: https://www.elastic.co/guide/en/elasticsearch/reference/current/keyword.html#mapping-unstructured-content

rw-access: yeah, definitely agree.  keyword gives you aggregations and wildcards. `text` is more for natural language and splits up words, punctuation, etc

generated/alerts/ecs/subset/malware_event/exceptionable.json
```diff
@@ -1,234 +1,4 @@
 [
-  "Endpoint.policy.applied.id",
```
rw-access, line 2: these changes intentional?

rw-access (approved):
i'm not sure about the other changes but `file.Ext.header_bytes` LGTM. thanks!

generated/alerts/ecs/subset/malware_event/exceptionable.json
AaronLamb1: no, they happened on "make".   @jonathan-buttner, is there someone else that should look at this to make sure this is not going to break everything?

custom_schemas/custom_file.yml
AaronLamb1: I don't know the answer to any of those, but I'll ask around.

generated/alerts/ecs/subset/malware_event/exceptionable.json
jonathan-buttner: I think these are auto generated by a tool in the chain. Not sure why they're being picked up now though 🤔

custom_schemas/custom_file.yml
rw-access: we didn't use `header_data` in Elasticsearch before, so we're fine.
for the endpoint-dev schema, we may want to keep in `header_data` in the schema even if it's not populated, so that  EQL rules can OR the two together:
```
(file.Ext.header_data[0] == 77 and file.Ext.header_data[1] == 95) or file.Ext.header_bytes : "MZ*")
```

AaronLamb1: nothing in detection-rules, but there are probably other teams that I don't even know exist, so probably safer to leave it in there.

AaronLamb1: hah, didn't see @rw-access comment, guess we can just leave it as is

generated/alerts/ecs/subset/malware_event/exceptionable.json
rw-access: i think pulling in master to this branch and regenrating might help

gabriellandau (approved):

jonathan-buttner (approved):
👍

PR #154, opened 2021-06-23
Add Process protection level to events and alerts

gabriellandau:
Issue: https://github.com/elastic/endpoint-dev/issues/8759

Windows has a concept of a Protected Process, which has strict code signing requirements and other limitations in order to keep malware out.

By adding process protection level to events and alerts, we can harden allowlist and trustedapp actions for known-problematic processes by adding a PPL requirement.   This rules out a variety of trivial backdoors that would otherwise be a concern.

This may help us avoid performance issues like: https://discuss.elastic.co/t/issues-with-elastic-agent-and-defender/269818

Note: Protection is a property of the process, not its access token.


matthewscherer (approved):
Was the code already added to the Endpoint?

bjmcnic (approved):

jonathan-buttner (approved):

PR #155, opened 2021-06-30
Add malware_signature for Memory protection alert

gogochan:
Issue: https://github.com/elastic/security-team/issues/1214

Adding additional fields that are part of yara signature. The fields are used in FileScore alert and MemoryScan Alert.

- [x] malware_signature.all_names
- [x] malware_signature.primary.signature.name
- [x] malware_signature.primary.signature.id
- [x] malware_signature.memory_region.allocation_protection
- [x] malware_signature.memory_region.allocation_size
- [x] malware_signature.memory_region.allocation_type
- [x] malware_signature.memory_region.region_protection
- [x] malware_signature.memory_region.region_size
- [x] malware_signature.memory_region.region_state
- [x] malware_signature.memory_region.memory_pe.imphash
- [x] malware_signature.memory_region.bytes_allocation_offset
- [x] ~malware_signature.primary.signature.hash.sha256~
- [x] ~malware_signature.primary.matches~
- [x] ~version~
- [x] ~identifier~


custom_subsets/elastic_endpoint/alerts/malware_event.yaml
```diff
@@ -469,6 +469,8 @@ fields:
                 exceptionable: true
               status:
                 exceptionable: true
+          malware_signature:
+            fields: "*"
```
gabriellandau, line 475: Everything else here looks like this:
```suggestion
            fields: {}
```

custom_subsets/elastic_endpoint/alerts/malware_event.yaml
```diff
@@ -469,6 +469,8 @@ fields:
                 exceptionable: true
               status:
                 exceptionable: true
+          malware_signature:
```
gabriellandau, line 474: Should this go into memory_protection_event, since Memory Scan is presented to the user as part of Memory Protection?  https://github.com/elastic/endpoint-package/blob/c23a052ef81ea39ab2934e2e1ea5b32bff102196/custom_subsets/elastic_endpoint/alerts/memory_protection_event.yaml

ferullo: This should stay where it is at the very least for malware (FileScore) events. We could duplicate it to another location as well though I'd vote against that.

custom_subsets/elastic_endpoint/alerts/malware_event.yaml
jonathan-buttner: We should update the docs on `"*"` vs `{}` because I always forget haha. `{}` only works when the field is a leaf aka if we only wanted `malware_signature.all_names` in the mapping we could write it like:

```
malware_signature:
  fields:
    all_names: {}
```
Since we want everything under `malware_signature` and `fields` is an array we have to use the `"*"` syntax like Chan has here.

custom_schemas/custom_malware_signature.yml
```diff
@@ -0,0 +1,40 @@
...
+      description: A sequence of signature names matched.
+
+    - name: identifier
+      level: custom
+      type: keyword
+      description: The model's unique identifier.
+
+    - name: primary
+      level: custom
+      type: object
```
jonathan-buttner, line 29: From the example the structure of this field is:

```
                "primary": {
                    "matches": [
                        "TWFpbEFjY291bnRDb25maWd1cmF0aW9u",
                        ...
                    ],
                    "signature": {
                        "hash": {
                            "sha256": "60c031526f8c3099f324b9dccaad3e8e7fb60c85ef79237aa9917e128b072c14"
                        },
                        "id": "d3ac2b2f-14fc-4851-8a57-41032e386aeb",
                        "name": "Windows.Trojan.AgentTesla"
                    }
                },
```

Do we want `matches` and the other fields to be in the mapping so they can be searched for? If yes, then we'll want to add them here as well.

You could so something like this:

```
    - name: primary.matches
      level: custom
      type: keyword
      description: some description

    - name: primary.signature.hash.sha256
      level: custom
      type: keyword
      description: some description

    - name: primary.signature.id
      level: custom
      type: keyword
      description: some description

    - name: primary.signature.name
      level: custom
      type: keyword
      description: some description
```

ECS core does have a hash schema: https://github.com/elastic/ecs/blob/1.7/schemas/hash.yml so if the endpoint would ever ship more hashes under `primary.signature.hash` (say md5 etc) we might want to change https://github.com/elastic/endpoint-package/blob/master/custom_schemas/custom_hash.yml to just overlay them onto here that way you don't have to duplicate them.

custom_subsets/elastic_endpoint/alerts/malware_event.yaml
ferullo: I take it back. I now see you're commenting on `process.Ext.malware_signature` not `file.Exe.malware_signature`.

Currently Linux only sends file activity based malware (FileScore) alerts (open, write, etc) so wouldn't ever need this location. But Windows and macOS can send process based malware alerts (process start). So I still _think_ this is needed for Windows and macOS signature based FileScore alerts (and wonder why testing has been OK without this so far).

But that's only whether or not this update is beneficial to malware alerts. My presumption would be that a standardized key name for yara signature matches regardless of the protection type would be better than it going in different places for different alert types. But I don't feel strongly, if my presumption is wrong I'll stay silent.

custom_schemas/custom_malware_signature.yml
```diff
@@ -0,0 +1,40 @@
...
+      description: The first matching details.
+
+    - name: secondary
+      level: custom
+      type: array
```
jonathan-buttner, line 34: In Elasticsearch there isn't an `array` type. Everything can be an array out of the box: https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping-types.html#types-array-handling

Could we change this to the structure and type that will be contained within `secondary`? If `secondary` will be the exact same as `primary`, you could abstract it out to a new file and have it overlaid as both `primary` and `secondary` using the ECS `reuse` mechanism.

jonathan-buttner (changes requested):
Thanks for the PR Chan! Couple questions and changes:

From the example that Kevin linked ([memory_signature_agentTesla.json.zip](https://github.com/elastic/security-team/files/6650515/memory_signature_agentTesla.json.zip)) it looks like `memory_region` is included in `malware_signature`: 

```
"memory_region": {
                    "allocation_base": 4194304,
                    "allocation_protection": "RWX",
                    "allocation_size": 245760,
                    "allocation_type": "PRIVATE",
                    "bytes_address": 4194304,
                    "bytes_allocation_offset": 0,
```

do we need to update https://github.com/elastic/endpoint-package/blob/master/custom_schemas/custom_memory_region.yml#L11 to include it within this schema as well?

custom_subsets/elastic_endpoint/alerts/malware_event.yaml
gogochan: Per our discussion, I will remove this line from this PR. We will add it back in 7.15

custom_schemas/custom_malware_signature.yml
gogochan: `secondary` will be a list of object that are subset of the `primary`. Should I just say `object` here then?

gogochan:
> do we need to update https://github.com/elastic/endpoint-package/blob/master/custom_schemas/custom_memory_region.yml#L11 to include it within this schema as well?

I'm not sure if we need to include them. IMO creating an Exception based on those values would be too much of information.

custom_schemas/custom_malware_signature.yml
jonathan-buttner: If we use `object` then users won't be able to search for it beyond just using a search based on its existence. If we don't expect users to search for the `secondary` fields then you could exclude it from the definition here. The only reason we want to define it in the endpoint package is so that users can search for it. If you want to include it just as a description of a field that could exist in an alert, we can add `enabled: false` (e.g. https://github.com/elastic/endpoint-package/blob/master/custom_schemas/custom_endpoint.yml#L28) that way Elasticsearch will not index it.

jonathan-buttner:
> > do we need to update https://github.com/elastic/endpoint-package/blob/master/custom_schemas/custom_memory_region.yml#L11 to include it within this schema as well?
> 
> I'm not sure if we need to include them. IMO creating an Exception based on those values would be too much of information.

I guess what I mean is, based on the example that Kevin gave I'm assuming the endpoint will send those memory region values. If we expect the user to want to search for alerts based on `malware_signature.memory_region.*` values then we'd need to include those in the mapping here. If we don't expect the user to search for them then we can leave them out. 

@kevinlog @pzl could you take a look at the mappings when you get a chance.

custom_schemas/custom_malware_signature.yml
```diff
@@ -0,0 +1,61 @@
...
+    - name: identifier
+      level: custom
+      type: keyword
+      description: The model's unique identifier.
+
+    - name: primary
+      level: custom
+      type: object
+      description: The first matching details.
+      fields:
```
jonathan-buttner, line 31: Let's remove the `fields:` here. That's only needed once at the top of these files.

kevinlog:
@jonathan-buttner thanks for bubbling this up.

> I guess what I mean is, based on the example that Kevin gave I'm assuming the endpoint will send those memory region values. If we expect the user to want to search for alerts based on `malware_signature.memory_region.*` values then we'd need to include those in the mapping here. If we don't expect the user to search for them then we can leave them out.

There are two questions to answer here:

1. Does the user need to search for alerts based on this field
2. Should the user be able to add Exceptions based on this field for this Alert
   - Implies we need it in the Index pattern

I don't think we need to index this field for searching.

However, we may want it included in the Index pattern so that we can write exceptions for it.

@gogochan @gabriellandau @ferullo any thoughts on writing Exceptions?

If we DO want to write Exceptions, we can `enabled: false` so that we don't index the field and include it in searches.

ferullo:
@joe-desimone has thoughts on what signature match fields should be available for exceptions.

joe-desimone:
I believe users should have the ability to search for and create exceptions from the following fields: `malware_signature.all_names`, `malware_signature.primary.signature.name`, and `malware_signature.primary.signature.id`, `malware_signature.memory_region.allocation_base`, `malware_signature.memory_region.allocation_protection`, `malware_signature.memory_region.allocation_size`, `malware_signature.memory_region.allocation_type`, `malware_signature.memory_region.region_protection`, `malware_signature.memory_region.region_size`, `malware_signature.memory_region.region_state`.

dstepanic17:
@joe-desimone Here, are a couple fields I did not see on your list.  I believe the `region_base` maps to `allocation_base`, but might be confusing for users to not let them search or create exceptions on it.  I believe the other field `bytes_allocation_offset` can bring in unique data so might be useful for searching/exceptions. Everything else looks good.

-  `malware_signature.memory_region.region_base`
- `malware_signature.memory_region.bytes_allocation_offset`

joe-desimone:
> @joe-desimone Here, are a couple fields I did not see on your list. I believe the `region_base` maps to `allocation_base`, but might be confusing for users to not let them search or create exceptions on it. I believe the other field `bytes_allocation_offset` can bring in unique data so might be useful for searching/exceptions. Everything else looks good.
> 
> * `malware_signature.memory_region.region_base`
> * `malware_signature.memory_region.bytes_allocation_offset`

Thank you. I actually had allocation_base on there by mistake, and left off region_base intentionally. My thinking was these are going to be generally random so not too helpful from an exceptionlist point of view (but useful for response). allocation_offset I waffled on. In some situations it could be useful so I added it now.

Also, we dont have it yet but we should add malware_signature.memory_region.memory_pe.imphash to match shellcode alert schema more closely. We would want to search+exception on that as well so I added it to my original list.

gogochan:
> I believe users should have the ability to search for and create exceptions from the following fields: malware_signature.all_names, malware_signature.primary.signature.name, malware_signature.primary.signature.id, malware_signature.memory_region.allocation_protection, malware_signature.memory_region.allocation_size, malware_signature.memory_region.allocation_type, malware_signature.memory_region.region_protection, malware_signature.memory_region.region_size, malware_signature.memory_region.region_state, malware_signature.memory_region.memory_pe.imphash, malware_signature.memory_region.bytes_allocation_offset

Just want to point out that the current workflow does exception list checking before reading  process memory, thus it lacks `memory_signature.*` fields.

https://github.com/elastic/endpoint-dev/blob/master/Elastic/Plugins/MemoryProtection/Lib/MemoryScan.cpp#L472-L479

We can alter the behavior, but it means that we are checking the exception list after scanning process memory.

gogochan:
> do we need to update https://github.com/elastic/endpoint-package/blob/master/custom_schemas/custom_memory_region.yml#L11 to include it within this schema as well?

@jonathan-buttner  We will make a separate PR to address this. I've adjusted the PR per comments provided. Let me know if this is good.

gabriellandau:
@joe-desimone per our recent conversations, do you think we should put `malware_signature` under `memory_region` instead of vice-versa?

custom_subsets/elastic_endpoint/alerts/malware_event.yaml
```diff
@@ -315,6 +315,8 @@ fields:
                       buffer: {}
                       decompressed_size: {}
                       encoding: {}
+          malware_signature:
+            fields: "*"
```
jonathan-buttner, line 319: The `memory_region` custom schema I referenced above defines more fields then were discussed in the ticket. Not a huge deal but if you want to limit it to only include the ones asked for you can specify them explicitly here in the form of:

```
malware_signature:
  fields:
    memory_region:
      fields:
        fields_a: {}
        field_b: {}
        ...
```

jonathan-buttner (changes requested):
Left a few more comments about changes I think we need

custom_schemas/custom_malware_signature.yml
```diff
@@ -0,0 +1,93 @@
...
+      level: custom
+      type: keyword
+      description: The id of the first yara rule matched.
+
+    - name: primary.signature.name
+      level: custom
+      type: keyword
+      description: The name of the first yara rule matched.
+
+    - name: malware_signature.memory_region.allocation_protection
```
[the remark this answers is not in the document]
gogochan: thanks... too many copy and pasting..

custom_schemas/custom_malware_signature.yml
```diff
@@ -0,0 +1,93 @@
...
+      level: custom
+      type: keyword
+      example: "RWX"
+      description: >
+        Original memory protection requested when the memory was allocated. Example values include "RWX" and "R-X".
+
+    - name: malware_signature.memory_region.allocation_size
```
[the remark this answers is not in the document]
gogochan: would it be okay even if fields defined under `memory_signature` is a subset of `memory_region`?

custom_subsets/elastic_endpoint/alerts/malware_event.yaml
gogochan: I see that you answered my previous question here.

gogochan: since memory scan is part of memory protection, I'll move this under `memory_protection_event.yaml` instead

gogochan:
Thanks @jonathan-buttner for the suggestion. It looks better now.

ferullo:
> We can alter the behavior, but it means that we are checking the exception list after scanning process memory.

Endpoint should definitely check the exception list after it scans memory and can generate the alert but before it takes action. Checking exceptions earlier than this is a good optimization but this final check is always needed for all protections.

custom_schemas/custom_malware_signature.yml
```diff
@@ -0,0 +1,41 @@
...
+  group: 2
+  short: These fields contain the yara malware signature.
+  description: >
+    These fields contain the yara malware signature.
+  reusable:
+    order: 1
+    top_level: false
+    expected:
+      - process.Ext
+      - file.Ext
```
jonathan-buttner, line 13: Just a heads up it doesn't look like `malware_signature` is being placed under `file.Ext` 🤔  that might be because the file subset file doesn't include it 🤷‍♂️  is that ok?

jonathan-buttner (approved):
Looks good! Just one question.

custom_schemas/custom_malware_signature.yml
gogochan: I updated the title. All fields added here are for Memory Protection plugin.

joe-desimone:
> > We can alter the behavior, but it means that we are checking the exception list after scanning process memory.
> 
> Endpoint should definitely check the exception list after it scans memory and can generate the alert but before it takes action. Checking exceptions earlier than this is a good optimization but this final check is always needed for all protections.

Agree here. I had the understanding that we do some early exceptionlist tests which would allow us better performance like https://github.com/elastic/global-security-exceptionlist/pull/257/files. Might be a good place for unit test.

joe-desimone:
> @joe-desimone per our recent conversations, do you think we should put `malware_signature` under `memory_region` instead of vice-versa?

Yes which would help makes things more consistent between shellcode alerts and memory scan alerts

joe-desimone:
> > @joe-desimone per our recent conversations, do you think we should put `malware_signature` under `memory_region` instead of vice-versa?
> 
> Yes which would help makes things more consistent between shellcode alerts and memory scan alerts

Created an issue to specifically discuss this proposal https://github.com/elastic/security-team/issues/1496

gogochan:
This may require further modification based on https://github.com/elastic/security-team/issues/1496

gogochan:
Added changes proposed in https://github.com/elastic/security-team/issues/1496

custom_schemas/custom_malware_signature.yml
```diff
@@ -0,0 +1,40 @@
...
+  title: Malware Signature
+  group: 2
+  short: These fields contain the yara malware signature.
+  description: >
+    These fields contain the yara malware signature.
+  reusable:
+    order: 1
+    top_level: false
+    expected:
+      - memory_region
```
gabriellandau, line 12: 👍

custom_subsets/elastic_endpoint/alerts/memory_protection_event.yaml
```diff
@@ -349,6 +349,27 @@ fields:
...
               user: {}
+          memory_region:
+            fields:
+              allocation_protection: {}
+              allocation_size: {}
+              allocation_type: {}
+              region_protection: {}
+              region_size: {}
+              region_state: {}
+              bytes_allocation_offset: {}
```
gabriellandau, line 360:
```suggestion
              bytes_allocation_offset: {}
              bytes_compressed: {}
              bytes_compressed_present: {}
              bytes_address: {}
              allocation_base: {}
```

gabriellandau (changes requested):

custom_subsets/elastic_endpoint/alerts/memory_protection_event.yaml
gogochan: do we want all of these to be searchable?

gogochan: or having them here doesn't impact searchability?

gabriellandau: I'm going to defer to @jonathan-buttner on this, but AFAIK they need them in the subset in order for them to end up in the resulting schema.  Otherwise, the presence of these fields is a schema violation.  We can disabling indexing with the `index: false` directive: https://github.com/elastic/endpoint-package/blob/d28668a39ab33c4c7ac620b6c6e0807fe0d9dcfd/custom_schemas/custom_memory_region.yml#L57-L60

gabriellandau (approved):
I pushed changes that addressed my concerns.  LGTM.

jonathan-buttner (changes requested):
Question about renaming `start_address_details`

custom_subsets/elastic_endpoint/alerts/memory_protection_event.yaml
```diff
@@ -262,27 +262,6 @@ fields:
               start_address_bytes: {}
               start_address_bytes_disasm: {}
               start_address_bytes_disasm_hash: {}
-              start_address_details:
```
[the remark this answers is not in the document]
gabriellandau: > Has the endpoint already started populating these fields in 7.14 or prior?

Only for diagnostic.  The shellcode feature isn't going live until 7.15.

> If it has, do any of our detection rules use these fields?

Negative.  Not until 7.15.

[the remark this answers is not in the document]
jonathan-buttner: Ok perfect! Thanks Gabe!

jonathan-buttner (approved):

PR #156, opened 2021-07-01
Bump ECS to 1.10

pzl:
Looking to update ECS to latest

This is targeted for `v0.21.0` package, so **after** the upcoming release

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 ROOT_DIR := $(shell dirname $(realpath $(firstword $(MAKEFILE_LIST))))
 # we are intentionally pinning the ECS version here, when ecs releases a new version 
 # we'll discuss whether we need to release a new package and bump the version here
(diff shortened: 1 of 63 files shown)


jonathan-buttner (commented):
The schema changes that came with the upgrade to v1.10.0 look fine. We should probably remove the generated template and component files (or maybe all of the generated files 🤷‍♂️ ) since it seems like the newer ECS tooling is creating a bunch of them now.

Maybe we should do that in a separate PR though? 

Basically we'd want to limit or remove what this line is copying:
`cp -r $(ROOT_DIR)/out/$(1)/generated/elasticsearch $(ROOT_DIR)/generated/$(1)`

jonathan-buttner (approved):

pzl:
> We should probably remove the generated template and component files (or maybe all of the generated files  ) 

Good point. Happy to do that in a followup PR

PR #167, opened 2021-08-03
Don't index memory region strings

gabriellandau:
It's a very-high-cardinality array.  Until we have a better solution, don't index them.


jonathan-buttner (approved):
1 line change results in 175 file changes 😢  we really need to fix these generated files haha.

joe-desimone (approved):

PR #172, opened 2021-08-12
new fields for improved call_stack details

jdu2600:
Adds additional thread information fields related to call stacks.
This improves our ability to write strong exceptionlist rules.

The additional fields are -
```
process.thread.Ext.call_stack_summary
process.thread.Ext.call_stack_final_user_module.name
process.thread.Ext.call_stack_final_user_module.path
process.thread.Ext.call_stack_final_user_module.hash.sha256
process.thread.Ext.call_stack_final_user_module.code_signature
```

### DIFF
--- a/custom_schemas/custom_process.yml
+++ b/custom_schemas/custom_process.yml
@@ -160,7 +160,7 @@
 
     - name: thread.Ext.call_stack_final_user_module.hash
       level: custom
-      type: nested
+      type: object
       description: Hashes of the call_stack_final_user_module.
 
     - name: thread.Ext.call_stack_final_user_module.hash.sha256
--- a/custom_schemas/custom_process.yml
+++ b/custom_schemas/custom_process.yml
@@ -89,44 +89,44 @@
       description: >
         Concatentation of the non-repeated modules in the call stack.
 
-    - name: thread.Ext.final_user_module
+    - name: thread.Ext.call_stack_final_user_module
       level: custom
       type: nested
       description: >
         The final non-win32 module in the call stack.
 
-    - name: thread.Ext.final_user_module.name
+    - name: thread.Ext.call_stack_final_user_module.name
       level: custom
       type: keyword
       example: "example.dll"
       description: >
-        The file name of the final_user_module.
+        The file name of the call_stack_final_user_module.
 
-    - name: thread.Ext.final_user_module.path
+    - name: thread.Ext.call_stack_final_user_module.path
       level: custom
       type: keyword
       example: "C:\\Program Files\\Example\\example.dll"
       description: >
-        The file path of the final_user_module.
+        The file path of the call_stack_final_user_module.
 
-    - name: thread.Ext.final_user_module.code_signature
+    - name: thread.Ext.call_stack_final_user_module.code_signature
       level: custom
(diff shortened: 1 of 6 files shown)


custom_schemas/custom_process.yml
```diff
@@ -82,6 +82,93 @@
       type: object
       description: Object for all custom defined fields to live in.
 
+    - name: thread.Ext.call_stack_summary
```
jonathan-buttner, line 85: Looks like we already have a `call_stack` custom schema here: https://github.com/elastic/endpoint-package/blob/master/custom_schemas/custom_call_stack.yml

Could we add these fields there? That custom schema is already overlaid at the location where these fields would be: `process.thread.Ext`: https://github.com/elastic/endpoint-package/blob/master/custom_schemas/custom_call_stack.yml#L12

If we can add them there, let's remove the prepended `call_stack_` part of the field names as it'd be redundant. So we'd have something like:

```
- name: stack_summary
...

- name: final_user_module
```

etc.

custom_schemas/custom_process.yml
```diff
@@ -82,6 +82,93 @@
...
+      description: >
+        Additional information about the certificate status.
+
+        This is useful for logging cryptographic errors with the certificate validity or trust status.
+        Leave unpopulated if the validity or trust of the certificate was unchecked.
+      example: ERROR_UNTRUSTED_ROOT
+
+    - name: thread.Ext.call_stack_final_user_module.hash
+      level: custom
+      type: nested
```
jonathan-buttner, line 163: Since `hash` only has a single child do we need it as `nested`? Do we plan to add additional hash values in the future and would we be expecting to query them such that we want to query the sha256 and say md5 at the same time?

I think we could probably get away with just using `type: object` here right?

custom_subsets/elastic_endpoint/alerts/memory_protection_event.yaml
```diff
@@ -204,6 +204,22 @@ fields:
...
+                    fields:
+                      exists: {}
+                      status: {}
+                      subject_name: {}
+                      trusted: {}
+                      valid: {}
+                  hash:
+                    fields:
+                      sha256:
+                        exceptionable: true
```
jonathan-buttner, line 222: The endpoint package doesn't use the `exceptionable` field anymore (we might still need to update a few places to document that 😬 ). So let's remove this field. If we're hoping to allow users to write exceptions based on certain fields we'll need to update these files:

https://github.com/elastic/kibana/tree/master/x-pack/plugins/security_solution/public/common/components/exceptions

`exceptionable_*` files.

custom_schemas/custom_process.yml
jonathan-buttner: Nested fields add some overhead, so we try to only use them when we need to. Here are some docs: https://www.elastic.co/guide/en/elasticsearch/reference/current/nested.html#nested-fields-array-objects

jdu2600:
Thanks for the advice and feedback.

re: testing
I forgot to mention that the changes pass endpoint-dev's EAF test suite which validates that all documents that endpoint produces match the endpoint_package+ecs schema.

PR #173, opened 2021-08-12
bump to ECS 1.11.0

pzl:
it's a big'n


jonathan-buttner:
Yikes 189 files. Could we try and remove some of the generated files from being checked in? I think we could ignore the elasticsearch directory. I haven't gotten around to it yet but we don't any of the `exception` stuff anymore and I think that'd remove the `subset` directories as well.

jonathan-buttner (approved):

PR #175, opened 2021-08-17
rule_detection_schema updates

brian-mckinney:
Schema updates for RulesEngine


custom_schemas/custom_responses.yml
```diff
@@ -0,0 +1,69 @@
+---
+- name: responses
```
ferullo, line 2: Should this be capitalized?

generated/alerts/elasticsearch/component/Endpoint.json
```diff
@@ -1,7 +1,7 @@
 {
   "_meta": {
     "documentation": "https://www.elastic.co/guide/en/ecs/current/ecs-Endpoint.html",
-    "ecs_version": "1.11.0"
+    "ecs_version": "1.10.0"
```
pzl, line 4: Looks like this needs to be regenerated

pzl (changes requested):
Looks like this needs to be regenerated. We updated to ECS-1.11. You may need to `rm -rf out/ecs` before `make`ing again, and that will ensure it pulls in the correct ECS version.

We're looking to fix that manual step so ECS updates aren't as much of a problem later

PR #180, opened 2021-08-19
[Maintenance] Remove 'exceptionable' field setting, and artifacts

pzl:
This is part of some package cleanup effort


paul-tavares (approved):
I'm 👍 from a perspective of "sure, looks ok" 😆 
But you should really get the 👍's from folks on the Detections side and Endpoint as this would be more of an impact (I think) for them

jonathan-buttner (approved):
Thanks for doing this @pzl !

ferullo:
I don't know how this PR works. As long as the same endpoint document fields are available for exceptions I'm happy.

pzl:
Who is a good person to have review and approve this from the perspective:

>  the same endpoint document fields are available for exceptions

marshallmain (approved):

ferullo:
> Who is a good person to have review and approve this from the perspective

I'm not sure. I guess @marshallmain and @jonathan-buttner approving means it must work? Functionally, a test would be to make sure that with this package in use you can still create malware alert exceptions on any field in the alert except the few we don't allow. For instance I am pretty sure `file.Ext.malware_classification.identifier` should be allowed but `file.Ext.quarantine_result` should not be, even though it is in the alert (since you Endpoint can't apply an exception after it has quarantined the file...)

jonathan-buttner:
> Who is a good person to have review and approve this from the perspective:
> 
> > the same endpoint document fields are available for exceptions

The changes look good. Nothing changed in the `package/endpoint` directory which is what matters for determining if the mapping is being changed.

pzl:
Off-thread, it appears Marshall was the original, intended consumer of the exceptionable files, and doesn't use them. I will take his sign-off as the deity-level-approval here

PR #183, opened 2021-08-23
[Maintenance] Remove pipenv, go back to basic virtualenv

pzl:
Biggest problem here was that pipenv didn't have or support a way to specify a *minimum minor version* of Python.

That is, we require python 3.6+. Python 3.7, 3.8, 3.9, etc. are all fine. But pipenv has no way to specify this logic. At best, it has `^3`, but that would include `3.0` through `3.5`, which don't work. In the pipenv ticket [requesting such support](https://github.com/pypa/pipenv/issues/1050) the authors have indicated no desire to add it.

We don't have terribly difficult setup here, just some packages that need to be installed. `requirements.txt` will probably work fine. And there is runtime python version checking inside the event generation script that exits if the python version is too low. That works for me


Makefile
```diff
@@ -122,22 +120,28 @@ $(REAL_ECS_DIR):
...
 setup-tools: $(ECS_TAG_REF)
-	pipenv install
 	git -C $(REAL_ECS_DIR) checkout $(ECS_GIT_REF)
-	cd $(REAL_ECS_DIR) && PIPENV_NO_INHERIT=1 pipenv --python 3.9 install -r scripts/requirements.txt
 
-gen-files: $(TARGETS) $(ESTC_PKG_BIN)
+$(VENV_DIR): $(VENV_DIR)/touchfile
+
+$(VENV_DIR)/touchfile: scripts/requirements.txt
+	test -d $(VENV_DIR) || python3 -m venv $(VENV_DIR)
```
jonathan-buttner, line 135: Just want to make sure I understand correctly, will this auto install the virtualenv type of environment in `VENV_DIR` as part of the build? Or will a user have to run this target manually?

jonathan-buttner (approved):

Makefile
pzl: correct, this will auto-set-up the virtualenv into that dir (which the default value for is `.gitignore`'d)

the places where python is used (`release` and the recipes under `%-target`), the directory is added as a dependency. The directory depends on the `touchfile` (easier for mod times), and touchfile's creation is here, that sets up everything.

So any change in `scripts/requirements.txt` should also trigger this `pip install` again (and retouch, updating the update-at time of the file).

So adding `$(VENV_DIR)` as a dependency should work somewhat "automagically"

pzl: N.B.:

in the actual running of the scripts, `pipenv run` is essentially replaced with `. venv/bin/activate; `, and the rest is the same.

PR #187, opened 2021-08-27
[Maintenance] Improve Makefile recipes 

pzl:
converts `Makefile` to track file relationships.

Specifically lists the expected end-result files as requirements to `all`, and then has all the rules needed to create the files (in the same places).

So things are still made into `out/` and then copied into `generated/` and then copied into `package/`. And schema files are still generated too.

This attempts to create all the same things in the same places, just with finer dependency tracking (so a subset change under `metrics` will just update _those_ files.). Since we can't easily track which _custom_schemas_ are used in which _custom_subsets_, any change in a custom_schema rebuilds it all. 

This also removes all the new ecs-generated subset things we don't need:
  - ecs/subset/*
  - elasticsearch/component/*

Further PRs to come to independently remove `schemas` and/or `generated/`, and needs (or lack of) can be discussed there


academo:
I don't know enough of the build process to evaluate if the changes are correct but the code makes sense to me. Extra kudos for adding comments to each section! :)

academo (approved):

Makefile
```diff
@@ -107,59 +69,91 @@ clean:
...
+
+# tags are omitted so they do not end up in .git/packed-refs. If we fetch separately, then they appear in .git/refs/tags/{}
 $(REAL_ECS_DIR):
 	git clone --no-tags https://github.com/elastic/ecs.git $@
 
 # deleting the tag helps deal with mod time differences between ecs dir and the tag ref file
 $(ECS_TAG_REF): $(REAL_ECS_DIR)
 	rm -rf $@
-	git -C $(REAL_ECS_DIR) fetch -t
+	git -C $(REAL_ECS_DIR) pull -t origin master
```
jonathan-buttner, line 131: What's the purpose of using pull and tracking the branch instead of fetch here? Do we need master to be able to checkout the pinned ecs version?

Makefile
```diff
@@ -94,11 +56,11 @@ endif
 SED := gsed
 endif
 
-.PHONY: all
-all:
-	$(MAKE) gen-files
 
-.PHONY: mac-deps
+all: $(PKG_FIELDS_TARGETS) $(DOC_TARGET) $(ESTC_PKG_BIN) $(SCHEMA_TARGETS) $(ECS_FLAT_TARGETS) $(ES_7_TARGETS)
```
jonathan-buttner, line 60: Are we still able to do make -j8?

Makefile
```diff
@@ -107,59 +69,91 @@ clean:
...
-
-.PHONY: check-docker
 check-docker:
 	docker -v || { echo "please install docker before running the package registry"; exit 1; }
 	docker-compose -v || { echo "please install docker-compose before running the package registry"; exit 1; }
 
-$(ROOT_DIR)/out:
-	mkdir -p $(ROOT_DIR)/out
+out:
+	mkdir -p $@
```
jonathan-buttner, line 149: This will create `out` in the current directory that the makefile is being run from right? I think in the clean target we're referencing `ROOT_DIR/out`. Do we want that to be just `out`?

Makefile
```diff
@@ -185,4 +179,4 @@ update-elastic-package:
 	go mod vendor
 
 # recipes / commands. Not necessarily targets to build
-.PHONY: update-elastic-package promote release lint run-registry clean
\ No newline at end of file
+.PHONY: all update-elastic-package promote release lint run-registry clean setup-tools mac-deps build-package check-docker package
```
jonathan-buttner, line 182: I might have missed it but I think `setup-tools` has been removed right?

jonathan-buttner (approved):
Left a few questions looks good though!

Makefile
pzl: So, I thought `fetch` would be sufficient originally, since it should just be grabbing new tag refs, and we are checking those out directly. But it turns out, we still need the `git fetch && git merge` for the new tags to work, and that combo is essentially a `pull`. 

And I added `origin master` just so it had a definitive target somewhere. 

Just fetching did give me an issue at some point and this seemed to work better

Makefile
pzl: Yes. And (perhaps) it can parallelize a bit better now, since the full dependency tree is spelled out

Makefile
pzl: I wasn't sure why there was so much absolute-pathing here. I don't know who's `make`ing this _not_ from the this directory (other than Endpoint team, maybe?)

Makefile
pzl: good catch, yes that's gone

PR #190, opened 2021-09-03
Better description for the "latest" transform

pzl:
Noticed this when adding the next transform. The existing transform might need some clarity. Open to other suggestions


joeypoon (approved):

pzl:
cc @kevinlog since this is technically, potentially user-facing copy (in transform management page)

academo (approved):

paul-tavares (approved):
🔥 🔥 🔥

PR #191, opened 2021-09-03
[United Meta] Add united transform, and current mapping

pzl:
This creates the baseline united transform, with mappings as we currently know them (copy of endpoint-metadata and copy of .fleet-agents nested under `united.endpoint` and `united.agent` respectively).

Screenshot of running transforms after package upgrade:

2021-09-03-153126_scrot

Screenshot of results (`GET /.metrics-endpoint.metadata_united_default/_search   {"query": {"match_all": {}}}`) 

2021-09-03-153700_scrot


joeypoon (approved):
🚀

PR #195, opened 2021-09-15
[Maintenance] Add some pull request templates

pzl:
Now, when people create PRs to the package, we have some reminders and requests for info, so we can match up the intent (e.g. use of the new schema field, and sample values) with the change and field type

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE/Schema_change.md
+++ b/.github/PULL_REQUEST_TEMPLATE/Schema_change.md
@@ -8,7 +8,7 @@ labels: 'schema, Team:Onboarding and Lifecycle Mgt'
 
 <!-- please describe the intended use for the field changes -->
 
-#### Sample values
+### Sample values
 
 <!--  Please provide sample values that will go into these field changes. 
     
--- a/.github/PULL_REQUEST_TEMPLATE/Schema_change.md
+++ b/.github/PULL_REQUEST_TEMPLATE/Schema_change.md
@@ -1,0 +1,35 @@
+---
+name: Schema Change
+about: Adds/Removes/changes any schema fields
+labels: 'schema, Team:Onboarding and Lifecycle Mgt'
+---
+
+## Change Summary
+
+<!-- please describe the intended use for the field changes -->
+
+#### Sample values
+
+<!--  Please provide sample values that will go into these field changes. 
+    
+    This ticket is a good reference: https://github.com/elastic/endpoint-dev/issues/9533
+-->
+
+Sample document:
+
+```json
+
+
+```
+
+
+## Release Target
+
+<!-- What is intended Kibana release this is expected to ship with -->
+
+
+## Q/A
+
+- [ ] I ran `make` after making the schema changes, and committed any generated files (in `schema/`, `generated/`)
+- [ ] Does this field need to be "exceptionable"? (no longer specified in this package, this is now tracked in Kibana)
+- [ ] If this is a `metadata` change, I also updated both transform destination schemas to match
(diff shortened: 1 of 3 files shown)


.github/PULL_REQUEST_TEMPLATE/Schema_change.md
```diff
@@ -0,0 +1,35 @@
...
+
+
+## Release Target
+
+<!-- What is intended Kibana release this is expected to ship with -->
+
+
+## Q/A
+
+- [ ] I ran `make` after making the schema changes, and committed any generated files (in `schema/`, `generated/`)
```
ashokaditya, line 33: Are we supposed to commit generated files as well?

.github/PULL_REQUEST_TEMPLATE/Schema_change.md
```diff
@@ -0,0 +1,35 @@
...
+- [ ] Does this field need to be "exceptionable"? (no longer specified in this package, this is now tracked in Kibana)
```
ashokaditya, line 34: What are "exceptional" fields?

.github/PULL_REQUEST_TEMPLATE/Schema_change.md
```diff
@@ -0,0 +1,35 @@
...
+- [ ] If this is a `metadata` change, I also updated both transform destination schemas to match
```
ashokaditya, line 35: There's only a single destination in the only transform so far, no? 🤔  https://github.com/elastic/endpoint-package/blob/f57b161ffe07ee6e2243c0a0d227a2673c1b27cf/package/endpoint/elasticsearch/transform/metadata_current/default.json#L13

ashokaditya (approved):
Nice. 🚀  I only have a few questions.

.github/PULL_REQUEST_TEMPLATE/Schema_change.md
pzl: For the time being, yes.

PRs may come down the line to _stop_ committing (and even generating) those two folders, and add to `.gitignore`. But until then, commit what gets `made`

.github/PULL_REQUEST_TEMPLATE/Schema_change.md
pzl: I believe these are fields in the `alerts` subset, for which a user is given a dropdown choice of fields to create alert-exceptions to.

.github/PULL_REQUEST_TEMPLATE/Schema_change.md
pzl: at this point in time, technically yes. 

That's going to change in the near future though. I didn't feel like rolling the language change from singular to "both transforms" in that other PR. But I can if you feel strongly

ashokaditya: Make sense

.github/PULL_REQUEST_TEMPLATE/Schema_change.md
ashokaditya: Ah ha!

ashokaditya: Also, not sure what it means that `this is now tracked in Kibana`

pzl: the place to specify that a field is exceptionable is in kibana. For people used to an old workflow where the used to add `exceptionable: true` to fields here, this line is supposed to inform them to go to kibana repo for that. 

Or rather, to let us know whether a field is intended to be something you can add exceptions on, and that we (OLM team) do that work in kibana.

PR #196, opened 2021-09-16
[Maintenance] Combine the PR templates into one

pzl:
_apparently_  I did the original one (https://github.com/elastic/endpoint-package/pull/195) wrong. You can do multiple templates like that, but there isn't a chooser screen like there is for Issues. 

A user would have to manually add `?template=Transform.md` to the URL when creating a PR to select a template. Not great.

And apparently the default was to show/populate nothing. Also not the intent.

So this combines it into one default for PRs.


ashokaditya (approved):
Nice!! 🚀

PR #197, opened 2021-09-16
Package spec update

pzl:
## Change Summary

Updating to latest (1.1.0) package spec version. And therefore updating the latest `elastic-package` tool

The latest package spec requires a changelog file, so this PR includes a changelog starting from GA release

There is no automation built around the version at the top of the changelog yet. We'll have to add something in the future to help there. The changelog is just going to be manual for a while

## Release Target

Kibana 7.16 / Package 1.2.0

---

This PR is huge. I tried to break the commits out, so they can be viewed individually, to ignore the package-tool update commit.

Perhaps we should just stop vendoring? cc @jonathan-buttner 

I don't know if doing that would put us into a *real* pickle if we had to go back and change things, on, say, the 7.15 branch. I don't think the newest `elastic-package` respects the fact that that was using `1.0.0` package spec format. It may still yell about the changelog. Doing work on past branches would require deleting your `elastic-package` binary, and re-grabbing it. Would it correctly grab an old version due to the `go.mod` alone? Of course, `vendor/` exists back in that history, but say we were in the future going back to _now_. Is `go.mod` sufficent to `go get` the right "past" elastic-package tool to validate?

Seems like it's going to be a headache no matter what, involving removing the tool and re-fetching. Might as well delete a gigantic directory along the way


jonathan-buttner:
I guess we'd only run into issues if one of the dependencies was deleted right? Doesn't look like `elastic-package` vendors anything so maybe we're fine with just the go module then.

pzl:
`elastic-package` is shipping pre-compiled binaries in github release pages now.

It doesn't cover _enough_ platforms to make we want to add fetching those in an automated fashion, but we should be able to un-stick ourselves with that at a minimum. And if we're in such a situation there's going to be _some_ manual intervention anyway. I don't see any make rules or scripting being bulletproof there

jonathan-buttner:
Gotcha, yeah deleting the vendor sounds good to me.

jonathan-buttner (approved):

Issue #211, opened 2021-11-04
Failing to generate Memory Protection alerts


nicholasberlin:
TL;DR- I think `process.Ext.memory_region.malware_signature.primary.matches` requires a different mapping.

We are seeing this error when trying to write memory protection alerts to elasticsearch
```
reason (object mapping for [process.Ext.memory_region.malware_signature.primary.matches] tried to parse field [null] as object, but found a concrete value) status (400)
```

This is the relevant json snippet of the alert
```
        "malware_signature": {
          "all_names": "Linux.EICAR.Not-a-virus",
          "identifier": "production-malware-signature-v1-linux",
          "primary": {
            "matches": [
              "WDVPIVAlQEFQWzRcUFpYNTQoUF4pN0NDKTd9JEVJQ0FSLVNUQU5EQVJELUFOVElWSVJVUy1URVNULUZJTEUhJEgrSCo="
            ],
...
        },
```

I think this is the current mapping:
```
"malware_signature" : {
                          "type" : "nested",
                          "properties" : {
                            "all_names" : {
                              "type" : "text"
                            },
                            "identifier" : {
                              "type" : "text"
                            },
                            "primary" : {
                              "type" : "nested",
                              "properties" : {
                                "matches" : {
                                  "type" : "nested"
                                },
                                "signature" : {
                                  "type" : "nested",
                                  "properties" : {
                                    "hash" : {
                                      "type" : "nested",
                                      "properties" : {
                                        "sha256" : {
                                          "type" : "keyword",
                                          "ignore_above" : 1024
                                        }
                                      }
                                    },
                                    "id" : {
                                      "type" : "keyword",
                                      "ignore_above" : 1024
                                    },
                                    "name" : {
                                      "type" : "keyword",
                                      "ignore_above" : 1024
                                    }
                                  }
                                }
                              }
                            },
                            "secondary" : {
                              "type" : "nested"
                            },
                            "version" : {
                              "type" : "keyword",
                              "ignore_above" : 1024
                            }
                          }
                        },
```

This is the full error message from the dev console:
```
{
  "error" : {
    "root_cause" : [
      {
        "type" : "mapper_parsing_exception",
        "reason" : "object mapping for [process.Ext.memory_region.malware_signature.primary.matches] tried to parse field [null] as object, but found a concrete value"
      }
    ],
    "type" : "mapper_parsing_exception",
[...]

PR #212, opened 2021-11-05
set malware_signature.primary.matches as unindexed keyword

pzl:
## Change Summary

Changes from `nested` to `keyword`. Looks like the expected values are an array of concrete values (strings, it looks like)

this stems from https://github.com/elastic/endpoint-package/issues/211

### Sample values

Sample document:

```json
        "malware_signature": {
          "all_names": "Linux.EICAR.Not-a-virus",
          "identifier": "production-malware-signature-v1-linux",
          "primary": {
            "matches": [
              "WDVPIVAlQEFQWzRcUFpYNTQoUF4pN0NDKTd9JEVJQ0FSLVNUQU5EQVJELUFOVElWSVJVUy1URVNULUZJTEUhJEgrSCo="
            ],
        },


```

## Release Target

7.16

### For mapping changes:

- [ ] ~~Does this field need to be "exceptionable"? (no longer specified in this package, this is now tracked in Kibana)~~
- [ ] ~~If this is a `metadata` change, I also updated both transform destination schemas to match~~

### DIFF
--- a/custom_schemas/custom_malware_signature.yml
+++ b/custom_schemas/custom_malware_signature.yml
@@ -38,6 +38,7 @@
       level: custom
       type: keyword
       description: The first matching details.
+      enabled: false
 
     - name: primary.signature.id
       level: custom
--- a/custom_schemas/custom_malware_signature.yml
+++ b/custom_schemas/custom_malware_signature.yml
@@ -36,7 +36,7 @@
 
     - name: primary.matches
       level: custom
-      type: nested
+      type: keyword
       description: The first matching details.
 
     - name: primary.signature.id
--- a/custom_schemas/custom_malware_signature.yml
+++ b/custom_schemas/custom_malware_signature.yml
@@ -38,7 +38,7 @@
       level: custom
       type: keyword
       description: The first matching details.
-      enabled: false
+      index: no
 
     - name: primary.signature.id
       level: custom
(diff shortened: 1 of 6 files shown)


gogochan:
Is there a way not to index it, but just keeping it within the store? This is just a binary blob, which I don't think we need to search.
https://github.com/elastic/endpoint-package/pull/155#issuecomment-872418477

pzl:
If we don't need to search it _or_ make exceptions on it, then we can just remove the mapping? or make it `enabled: false` I believe.

If a field is just incidentally retrieved with the rest of a document, we don't need to map it at all.

If we need to search, filter, sort, alert, or alert-exception on a field, we generally need to map it. Do we know what the intent behind adding this field's mapping was? IIRC it was part of a large group of additions in #200

joeypoon (approved):

ferullo:
@joe-desimone @dstepanic17 @threat-punter do you have thoughts on Dan's question about exceptions?

joe-desimone:
I have always thought of matches as informational, so I think not searchable or exceptionable would be fine. Unless Dan S/Seth have a specific use case I am not thinking of.

dstepanic17:
I can't think any of use cases that come to mind. There should be other fields that are better for exceptionlisting and they also don't need to be searchable.

gogochan (approved):

kevinlog (approved):

PR #218, opened 2021-11-18
Set both transforms to be "managed" mode

pzl:
## Change Summary

Adds `Managed` flag to both transforms, marking them as "system" transforms

2021-11-18-141032_scrot

## Release Target

8.0.0 (package `1.3` series)

### For mapping changes:

- [X] I ran `make` after making the schema changes, and committed any generated files (in `schema/`, `generated/`)
- [ ] ~~Does this field need to be "exceptionable"? (no longer specified in this package, this is now tracked in Kibana)~~
- [ ] ~~If this is a `metadata` change, I also updated both transform destination schemas to match~~

### For Transform changes:

- [ ] ~~The corresponding transform destination schema was updated if necessary~~

### DIFF
--- a/package/endpoint/elasticsearch/transform/metadata_current/default.json
+++ b/package/endpoint/elasticsearch/transform/metadata_current/default.json
@@ -19,6 +19,9 @@
         "sort": "@timestamp"
     },
     "description": "Latest Endpoint metadata document per host",
+    "_meta": {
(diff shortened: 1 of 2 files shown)


ashokaditya (approved):
🚀 🐑

academo (approved):

PR #221, opened 2021-12-03
Make 1.3 series target 7.17 release

pzl:
## Change Summary

With the addition of a 7.17 stack, let's increment the package to follow.

package `1.2.x` series will continue to serve stack `7.16`.

package `1.3.x` series will target 7.17 stack

which will bump `1.4.x` package series to track kibana `8.0`

### DIFF
--- a/package/endpoint/manifest.yml
+++ b/package/endpoint/manifest.yml
@@ -16,7 +16,7 @@ policy_templates:
     description: Interact with the endpoint.
     multiple: false
 conditions:
-  kibana.version: "^8.0.0"
+  kibana.version: "^7.17.0"
   # See https://github.com/Masterminds/semver#caret-range-comparisons-major for more details on `^` and supported versioning
   # >= <the version> && < 8.0.0
 icons:


kevinlog:
@pzl just so we're on the same page - will the `7.17` package effectively be the same as the `7.16` package for now?  Just wondering about any changes that went out in `8.0` that shouldn't go out in `7.17`.  For instance, I'm not sure if the `managed` transform tag is implemented in the `7.x` series.

pzl:
I didn't make any changes to strip out things, no. I think `managed` is the only difference between current git tip and `7.16` release.

The `managed` should be ignored in `7.17`. It's in an arbitrary key/value metadata section

kevinlog:
okay - but with the new version line, be able to release another `7.17` package on it's own line if we ran into a problem and not affect `8.0.0`.  do I have that right?

pzl:
correct. We can individually target a stack release without affecting others. Selectively backport (if needed) without being _forced_ to cascade forward

kevinlog (approved):

PR #223, opened 2021-12-20
Hide .metrics-metadata-united

pzl:
## Change Summary

Starting in ES 8.0, hidden indices do not just mean start-with-leading-dot. The `index.hidden` setting was added.

This sets our dot-prefixed united metadata index to have `hidden: true`

## Release Target

8.0

### For mapping changes:

### For Transform changes:


kevinlog (approved):
LGTM

PR #228, opened 2022-01-28
add code signature fields for macOS signatures

intxgo:
## Change Summary

I was recently working on adding more signature information to documents in order to improve detection, etc.

We added `code_signature` information to any document leaf containing info about a binary.

There are two entities, their use depends on the OS.
1. single code signature `[dotted path].code_signature`
```
"process": {
    "code_signature": {
        "signing_id": "com.apple.bash",
        "trusted": true,
        "subject_name": "Software Signing",
        "exists": true,
        "team_id": "",
        "status": "No error."
    },
    "executable": "/bin/bash"
},
```
2. nested array of code signatures `[dotted path].Ext.code_signature`
This array is sent only from Windows, thus the code signatures in the array will never have `team_id` and `signing_id`. Furthermore the first signature (also called the primary authenticode signature) from the array is repeated for compatibility reasons with ECS above the `Ext` path, however likewise Endpoint will never send `team_id` and `signing_id` fields in the primary signature.
```
"process": {
    "Ext": {
        "code_signature": [
            {
                "trusted": true,
                "subject_name": "Microsoft Corporation",
                "exists": true,
                "status": "trusted"
            }
        ]
    },
    "code_signature": {
        "trusted": true,
        "subject_name": "Microsoft Corporation",
        "exists": true,
        "status": "trusted"
    }
}
```
Historically Endpoint fetched only the primary authenticode signature on Windows, thus the array under `Ext` leaf was added later when we enhanced Endpoint to fetch all authenticode signatures and we didn't want to make any breaking changes in the schema. Recently added documents did't have the explicit placeholder for primary authenticode signature, only the array, however now as we are introducing macOS signatures it makes sense to unify all documents. MacOS does not support multiple signatures.

This PR doesn't contain yet files generated by `make`. However I run `make` locally, it only generated a new README for me. I'm not sure if this is expected. The README looks good.

### Sample values

Sample document:

```json


```

### For mapping changes:

- [ ] I ran `make` after making the schema changes, and committed any generated files (in `schema/`, `generated/`)

### For Transform changes:


kevinlog:
@pzl fyi - is it possible to run this using the test harness?  We could point @intxgo to the documentation on how to do it

pzl:
@intxgo is this intended to ship for kibana `8.1` or `8.2`?

intxgo:
@pzl I don't know, endpoint 8.1 sends all of that, but some fields were send even before.

pzl:
this can be backported to wherever it's needed. But I wouldn't count needed as just "endpoint sending that field". But where we may actually use detections, filtering, etc. Do we have any detections relying on this that will ship in 8.1 (or shipp*ed* in 8.0?)  Is there maybe nothing still using this yet, and it was a forward-looking enhancement for the future?

intxgo:
We should ask around in the protections team, as they were the driving force for this change. For endpoint, it's just all about test automation. We validate if documents sent to Elasticsearch conform to ECS and endpoint-package, currently they don't so we have to keep exceptions https://github.com/elastic/endpoint-dev/blob/main/ElasticCommonSchema/schema_violations to make our Jenkins runs GREEN

joe-desimone:
For protections, we should be able to ship logic which will apply to 8.0+ endpoints. I don't think that has happened yet but @brokensound77 @DefSecSentinel @Samirbous would know. 

This field is useful beyond our out of the box protections. Very useful for event filtering, exception lists, and for users hunting across their data as well. So if backporting to 8.0 is easy enough we should probably do it.

pzl:
backporting to 8.0 is likely possible. We will test for it

pzl (approved):
this is ok to go in, but you need to merge in recent changes and regenerate

intxgo:
merged changed from master and added make generated files

PR #234, opened 2022-03-09
Ransomware Data Type Changes

magermark:
## Change Summary

DO NOT REVIEW!

### Sample values

Sample document:

```json


```

### For mapping changes:

- [ ] I ran `make` after making the schema changes, and committed any generated files (in `schema/`, `generated/`)

### For Transform changes:


ferullo:
Does this need to be backported to 7.17?

magermark:
@ferullo Yes, this would need to be backported to `7.17` to ensure a consistent user experience across the board

crowens:
Should we consider this a breaking change?  @andrewkroh what do we usually do in a package when we change a field type? @bradenlpreston

andrewkroh:
When the integration is upgraded by Fleet the data stream will be rolled over in order to apply the new mapping. Any data that is ingested after that will need to align to the type (keyword). In this case that won't be a problem since binary data must be base64 encoded and base64 is obviously valid for a keyword field.

You may want to add a step to your Ingest Node pipelines to conditionally convert incoming base64 data to hexidecimal to account for Agents that are not yet upgraded. I think you'd need to use Painless to handle this (like https://www.elastic.co/guide/en/elasticsearch/painless/master/painless-api-reference-shared-java-util.html#painless-api-reference-shared-Base64-Decoder).

The last issue to consider are queries. Are they any consumers (rules, alerts, dashboards, etc) of this field in the Elastic ecosystem that will be broken by the change or by receiving mixed results (base64 and hex) when hitting older indices? You should test this on a cluster having both a binary and keyword mapped indices to see if you encounter any other issues (like check Discover and check the Kibana index pattern / data view UI).

magermark:
Bumping for increased visibility...

I'm going to begin testing this today to validate it operates as expected, but would appreciate review feedback just to make sure the content of the PR makes sense.

CC: @pzl @joeypoon

pzl (approved):
the change here is ok

sounds like what's being sent from endpoint is not changing at all, just how it's stored, so that obviates some of the above complications.

This won't re-index the existing data, it will just index anything after package update. Not much to do there.

If this needs to be exceptionable, then it probably needs to be added to the Kibana repo here: `x-pack/plugins/security_solution/public/common/components/exceptions/exceptionable_endpoint_fields.json`

joeypoon (approved):

PR #239, opened 2022-03-18
Linux event model (for session_view plugin)

mitodrummer:
## Change Summary

The 8.2 release will include the following update to ECS
https://github.com/elastic/ecs/blob/main/rfcs/text/0030-linux-event-model.md

This PR is to add the mappings to the logs-endpoint process and alerts indices to allow for the new Session View feature to work in 8.2.

I believe some of the changes here will need to be cleaned up once 8.2 ECS is released.

## Release Target

8.2

### For mapping changes:

### For Transform changes:


custom_schemas/custom_process.yml
```diff
@@ -347,3 +356,68 @@
...
+      short: The TTY character device's major number.
+      description: >
+        The major number identifies the driver associated with the device. The character device's major and minor numbers can be algorithmically combined to produce the more fam
+
+    - name: tty.char_device.minor
+      level: extended
+      type: long
+      short: The TTY character device's minor number.
+      description: >
+        The minor number is used only by the driver specified by the major number; other parts of the kernel don’t use it, and merely pass it along to the driver. It is common f
```
opauloh, line 422: it seems the text was cut

generated/alerts/ecs/ecs_flat.yml
```diff
@@ -8445,677 +8469,2525 @@ process.entity_id:
...
-process.name:
-  dashed_name: process-name
+process.entry_leader.interactive:
+  dashed_name: process-entry-leader-interactive
+  description: 'Whether the process is connected to an interactive shell. Process
+    interactivity is inferred from the processes file descriptors. If the character
+    device for the controlling tty is the same as stdin and stderr for the process,
+    t Note: A non-interactive process can belong to an interactive session and is
+    simply one that does not have open file descriptors reading the controlling TTY
+    on FD 0 (stdi'
```
opauloh, line 8621: stdin?

generated/alerts/ecs/ecs_flat.yml
```diff
@@ -8445,677 +8469,2525 @@ process.entity_id:
...
-  level: custom
-  name: Ext.code_signature.trusted
+process.entry_leader.same_as_process:
+  dashed_name: process-entry-leader-same-as-process
+  description: 'This boolean is used to identify if a leader process is the same as
+    the top level process. For example, if process.group_leader.same_as_process =
+    true, it means the process event in question is the leader of its process group.
+    Details under process.* like pid The same applies for both process.session_leader
+    and process.entry_leader. This field exists to the benefit of EQL and other rule
+    engines since it''s not possible to compare equality between two fields in a single
```
opauloh, line 8808: `it''s`

mitodrummer:
Also, it seems people are getting alerts written to the logs-endpoint.alerts index? cc @kqualters-elastic I'm not seeing this index locally when creating rules for endpoint. Is there anything i'm missing. I notice there is a ticket about session data not appearing in the endpoint alerts index. Could that be related?

mitodrummer:
image

Was able to trigger a malware alert, which created the index (thx @kqualters-elastic for the help). Looks like the mappings are applied.

It sounds like other indices like .alerts-security and .alerts-default depend on the ECS 8.2 RFC getting merged though, so this PR will only unblock part of the story. Also, I noticed session data is not set in alerts written to logs-endpoint.alerts, but it seems that is a known issue https://github.com/elastic/endpoint-dev/issues/10692

generated/alerts/ecs/ecs_flat.yml
mitodrummer: i think this is an escape for the quote, as part of the generated files.

kqualters-elastic (approved):
ya think the point of this is just to ensure that the new fields are in the logs-endpoint mapping, and they appear to be, lgtm 👍

custom_schemas/custom_host.yml
```diff
@@ -0,0 +1,21 @@
+---
+- name: host
+  title: Custom Host
+  group: 2
+  short: To be replaced by 8.2 ecs mappings
+  description: >
+    To be replaced by 8.2 ecs mappings
+  type: group
```
pzl, line 8: I think the way this is specified is overriding some existing `Host` data

package/endpoint/data_stream/metadata/fields/fields.yml
```diff
@@ -319,11 +319,9 @@
 
         This field is an array. This will allow proper categorization of some events that fall in multiple event types.'
 - name: host
-  title: Host
+  title: Custom Host
   group: 2
-  description: 'A host is defined as a general computing instance.
-
-    ECS host.* fields should be populated with details about the host on which the event happened, or from which the measurement was taken. Host types include hardware, virtual machines, Docker containers, and Kubernetes nodes.'
+  description: To be replaced by 8.2 ecs mappings
```
pzl, line 324: an example of where it's being overridden where we probably don't need it to be

mitodrummer: I see, does omiting those properties fix this override issue?

mitodrummer: Or should i just pull the ECS versions of these things?

pzl: Perhaps mirroring the title and description will end up with a no-op for these parts, since we will still need to add `host.boot.id` and the other `host.` field, as they don't exist in ECS 1.11

mitodrummer: removing description allowed the ecs one to funnel through. title is fixed. should be g2g

pzl (approved):
Changes are good from my perspective, as long as this generates the mappings you expect in endpoint alerts & process indices

PR #240, opened 2022-03-22
Added missing mapping for process.env_vars

mitodrummer:
## Change Summary

Small amendment to the previous linux-process-model PR.

## Release Target

8.2

### For mapping changes:


pzl (approved):
quick n easy. Looks good to me

PR #243, opened 2022-03-31
[Security Solution] Increase delay of united transform

kevinlog:
## Change Summary

In Endpoint management dev testing, we noticed that the shorter delays in the united transform were periodically dropping documents when using our endpoint generator dev tool.

When I was running tests which added ~30 Agent and Endpoint docs resulted in the `united` transform, I noticed that documents from our generator dev tool would periodically be dropped.  I believe this has to do with look back in the transform when combining the Agent and Endpoint documents.

After increasing the delay in the `united` transform, all transforms kept up without problems when I left the generator running, adding a total of 2000 Agent and Endpoint docs.

See the Agent/Endpoint counts match below.

image

image

## Release Target

`8.2`

### For Transform changes:

### DIFF
--- a/package/endpoint/elasticsearch/transform/metadata_united/default.json
+++ b/package/endpoint/elasticsearch/transform/metadata_united/default.json
@@ -11,7 +11,7 @@
     "frequency": "1s",
     "sync": {
         "time": {
-            "delay": "1s",
+            "delay": "5s",
             "field": "updated_at"
         }
     },
--- a/package/endpoint/elasticsearch/transform/metadata_united/default.json
+++ b/package/endpoint/elasticsearch/transform/metadata_united/default.json
@@ -11,7 +11,7 @@
     "frequency": "1s",
     "sync": {
         "time": {
-            "delay": "5s",
+            "delay": "4s",
             "field": "updated_at"
         }
     },


kevinlog:
@pzl thanks, I lowered to 4s to be consistent with the prior value.

>I think this might point towards an issue when agent+endpoint docs are added in bulk in rapid succession, but I'm not sure about the real-world case bumping up against this. Maybe if an entire office booted up and enrolled at precisely the same time, but even that seems unlikely.

I agree that what the generator does is an extreme case and may not be likely in the real world, but we still can't predict when documents will be ingested so I'd rather err on the side of caution and be comfortable with that extreme case.  We're still getting a good UX performance boost that I noticed while testing.  Thoughts?

cc @joeypoon

pzl (approved):
Not going to stand in the way of better-safe-than-sorry, but I will open a ticket to investigate getting back to 1-2s, to see if we can ever reproduce outside of a generator

joeypoon:
maybe we should do some local horde testing with a few thousand to see if it's more of a generator problem? I do find it really strange that during our screen share yesterday one of the missing docs had only a .2 second difference between agent and endpoint doc. but I do agree with Dan that there's no harm in being safer with this for now.

Issue #245, opened 2022-04-05
Unmapped fields


pzl:
In setting up automated testing for the soon to come CI runs, there is a backlog of failures to clear to get to green. 

In running static tests, I grabbed sample documents from `_source` from each data stream from real endpoints to use. The static test flags any fields that are present that _do not_ have a mapping (among other issues like data type mismatch).

Here are the outstanding issues from unmapped fields:

<details>

  <summary>See log output</summary>

  ```
Run static tests for the package
--- Test results for package: endpoint - START ---
FAILURE DETAILS:
endpoint/action_responses Verify sample_event.json:
[0] field "EndpointActions.input_type" is undefined
endpoint/file Verify sample_event.json:
[0] field "elastic.agent.id" is undefined
[1] field "process.Ext.code_signature" is undefined
[2] field "process.parent.pid" is undefined
[3] field "process.code_signature.status" is undefined
[4] field "process.code_signature.trusted" is undefined
[5] field "process.code_signature.subject_name" is undefined
[6] field "process.code_signature.exists" is undefined
endpoint/library Verify sample_event.json:
[0] field "elastic.agent.id" is undefined
endpoint/metrics Verify sample_event.json:
[0] field "Endpoint.metrics.malicious_behavior_rules" is undefined
[1] field "Endpoint.metrics.cpu.endpoint.histogram.counts" is undefined
[2] field "Endpoint.metrics.cpu.endpoint.histogram.values" is undefined
[3] field "elastic.agent.id" is undefined
endpoint/network Verify sample_event.json:
[0] field "process.Ext.code_signature" is undefined
[1] field "process.code_signature.status" is undefined
[2] field "process.code_signature.trusted" is undefined
[3] field "process.code_signature.subject_name" is undefined
[4] field "process.code_signature.exists" is undefined
[5] field "elastic.agent.id" is undefined
endpoint/policy Verify sample_event.json:
[0] field "elastic.agent.id" is undefined
endpoint/process Verify sample_event.json:
[0] field "elastic.agent.id" is undefined
endpoint/registry Verify sample_event.json:
[0] field "elastic.agent.id" is undefined
[1] field "process.Ext.code_signature" is undefined
[2] field "process.code_signature.trusted" is undefined
[3] field "process.code_signature.subject_name" is undefined
[4] field "process.code_signature.exists" is undefined
[5] field "process.code_signature.status" is undefined
endpoint/security Verify sample_event.json:
[...]

PR #246, opened 2022-04-05
Tests

pzl:
## Change Summary

Support the automated tests

This adds:
- pipeline tests to verify our ingest pipelines work as intended (all but 1 are very basic), but this does include testing for proper geo-ip enrichment
- static tests that verify our mapping against documents pulled from Endpoints

As discussed in https://github.com/elastic/endpoint-package/issues/245 , to get the mapping tests passing, I added the `code_signature` mappings to several additional indices. That is the primary mapping change here

Documentation to come in a follow up PR with some CI adjustments after this


package/endpoint/data_stream/action_responses/_dev/test/pipeline/test-ingest-timestamp.json-expected.json
```diff
@@ -0,0 +1,9 @@
+{
+    "expected": [
+        {
+            "event": {
+                "ingested": "2022-04-03T19:44:49.026381581Z"
```
andrewkroh, line 5: As of 7.14 none of the data streams are required to set `event.ingested` because Fleet does this automatically in a final pipeline. You can remove this from all of your pipelines which will simplify these tests and make reviewing future changes easier because this always-changing timestamp will not appear in future diffs.

You can also remove the `dynamic_fields` test config.

package/endpoint/data_stream/file/fields/fields.yml
```diff
@@ -908,6 +908,49 @@
       ignore_above: 1024
       description: An array of entity_ids indicating the ancestors for this event
       default_field: false
+    - name: Ext.code_signature
+      level: custom
```
andrewkroh, line 912: Many of the properties included into the definition are unnecessary. I recommend only including the ones mentioned in the package-spec (https://github.com/elastic/package-spec/blob/d0173302e6b2435556abc060d67f1e10d232f2bd/versions/1/data_stream/fields/fields.spec.yml) so that in the future when this file is strictly validated this will be in good shape. Properties like `level`, `example`, `default_field` are not part of the spec and are not utilized by Fleet.

andrewkroh (commented):
Added some general recommendations (not necessarily specific to this PR).

PR #255, opened 2022-04-26
upgrade ECS to 8.1

joeypoon:
## Change Summary

Upgrade ECS `1.11.0` -> `8.1.0`

### For mapping changes:

### DIFF
--- a/package/endpoint/data_stream/alerts/fields/fields.yml
+++ b/package/endpoint/data_stream/alerts/fields/fields.yml
@@ -10,25 +10,27 @@
 
     Required field for all events.'
   example: '2016-05-23T08:05:34.853Z'
+  default_field: true
 - name: Events
   level: custom
   type: object
   description: events array
-  default_field: false
 - name: message
   level: core
-  type: text
+  type: match_only_text
   description: 'For log events the message field contains the log message, optimized for viewing in a log viewer.
 
     For structured logs without an original message field, other fields can be concatenated to form a human-readable summary of the event.
 
     If multiple messages exist, they can be combined into one message.'
   example: Hello World
+  default_field: true
 - name: Endpoint
   title: Endpoint
   group: 2
   description: Fields describing the state of the Elastic Endpoint when an event occurs.
   type: group
+  default_field: true
   fields:
     - name: policy
       level: custom
@@ -131,6 +133,7 @@
   group: 2
   description: These fields contain information specific to Memory Protection alerts.
   type: group
+  default_field: true
   fields:
     - name: cross_session
       level: custom
@@ -170,6 +173,7 @@
   group: 2
   description: These fields contain information specific to ransomware alerts.
   type: group
+  default_field: true
   fields:
     - name: child_processes.executable
       level: custom
@@ -360,6 +364,7 @@
   group: 2
   description: responses array contains rule response action results
   type: group
+  default_field: true
   fields:
     - name: '@timestamp'
       level: custom
@@ -426,6 +431,7 @@
 
     These fields provide more context about the target process and thread that are related to the data in the document. Useful in a security context where a target process or thread may be acted on by another process or thread.'
   type: group
+  default_field: true
   fields:
     - name: dll.Ext
       level: custom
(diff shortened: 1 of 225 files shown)


package/endpoint/data_stream/alerts/fields/fields.yml
```diff
@@ -810,6 +831,15 @@
...
       default_field: false
+    - name: process.Ext.dll.code_signature.digest_algorithm
+      level: extended
+      type: keyword
+      ignore_above: 1024
+      description: 'The hashing algorithm used to sign the process.
+
+        This value can distinguish signatures when a file is signed multiple times by the same signer but with a different digest algorithm.'
+      example: sha256
+      default_field: false
```
pzl, line 842: I think we want to adjust our custom_subsets here to not pull in these new fields. Alerts is already too large, and I don't think Endpoint is using this field. the `"*"` may be contributing some bloat for us during ECS upgrades

package/endpoint/data_stream/alerts/fields/fields.yml
```diff
@@ -2099,13 +2178,6 @@
       description: Process id.
       example: 4242
       default_field: false
-    - name: process.parent.ppid
-      level: extended
-      type: long
-      format: string
-      description: Parent process' pid.
-      example: 4241
-      default_field: false
```
pzl, line 2108: I don't think we want to actually _lose_ this field on our output. Endpoint may be sending this still

If Endpoint is not sending `process.parent.ppid` or `process.ppid` at all for any alerts, then I guess we can proceed with removal. But I suspect they might use it. In which case consult the ECS guide for where this information is supposed to be in a perfect-ecs-following world. Maybe it was renamed to go somewhere else? In that case, we could change `process.parent.ppid` to be an `alias` type pointing to the new place.

Or, if there is no guidance for what this was renamed to or where it goes, then we just need to re-add it as a custom field so we don't actually lose it in the data stream (again, assuming we confirm with Endpoint team that they do use the `ppid` fields).

package/endpoint/data_stream/alerts/fields/fields.yml
```diff
@@ -4969,8 +4996,7 @@
       default_field: false
     - name: entry_leader.command_line
       level: extended
-      type: keyword
-      ignore_above: 1024
+      type: wildcard
```
pzl, line 4999: do we know what the implication here of `keyword`-->`wildcard` is?

package/endpoint/data_stream/alerts/fields/fields.yml
```diff
@@ -8072,6 +8138,12 @@
       ignore_above: 1024
       description: SHA256 hash.
       default_field: false
+    - name: enrichments.indicator.file.hash.sha384
+      level: extended
+      type: keyword
```
pzl, line 8143: similar to all the things popping up under `code_signature` everywhere -- I don't think we want all these extra enrichments, unless explicitly asked by Endpoint team. Probably another mistake to have `"*"` in the subset files

pzl (commented):
Nice to see it builds. Now we need to work around 

- determining whether any removed field (user, ppid, some hashes) is actually used/sent by Endpoint, and keeping it around via aliases or as a custom field
- removing some of our `"*"` usage that is pulling in too many fields

and then probably battling with the tests once thats over

joeypoon:
@pzl ready for another look:
* got rid of the new fields being pulled in by the `"*"`s.
* since `keyword` -> `wildcard` is happening at ECS definition level, I'm inclined to leave them as whatever ECS defaults them as. there's no functional difference, only performance depending on the usage and I imagine they spent more time than I have determining the best default.
* there are some removed `pe/hash.*` fields since they were moved to `threat.enrichments.indicator.file.pe/hash.*`. I'm waiting on Ferullo to confirm whether these are in use but I don't see them being populated in the `logs-endpoint.alerts-default` index so I believe these are fine to leave out.

pzl (commented):
Looks pretty good. I think the only thing remaining is all the `enrichment` and `indicator` stuff. Everything else looked :ok_hand:

pzl (approved):
:+1:

PR #258, opened 2022-05-11
bump kibana.version > ^8.3

joeypoon:
## Change Summary

bump kibana.version to ^8.3


pzl (approved):
thank you. let's see how CI handles this. It affects what stack it tries to run. might be a small fix there if it borks

PR #265, opened 2022-06-21
Add more Endpoint fields

ferullo:
## Change Summary

This updates mappings to account for a number of fields Endpoint already generates.

I'm not sure if all the fields added need to be indexed or if the types are right. Some back and forth/thorough review would be appreciated.

### Sample values

There are too many changes to easily share a sample document, though I can get some samples if really needed. The changes in this PR pass Endpoint testing.

## Release Target

Whenever.

## Q/A

Endpoint automated testing passes with these changes and exceptions that allowed testing to otherwise pass removed.

### For mapping changes:

### For Transform changes:


custom_schemas/custom_process.yml
```diff
@@ -361,6 +361,62 @@
         Indicates the protection level of this process.  Uses the same syntax as Process Explorer.
         Examples include PsProtectedSignerWinTcb, PsProtectedSignerWinTcb-Light, and PsProtectedSignerWindows-Light.
 
+    - name: Ext.device.bus_type
```
ferullo, line 364: Will copy paste here too.

custom_schemas/custom_file.yml
```diff
@@ -287,6 +287,48 @@
       description: >
         Primary malware signature version.
 
+    - name: Ext.device.bus_type
+      level: custom
+      type: keyword
+      short: FILL ME IN
```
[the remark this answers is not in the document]
gabriellandau: FYI you can use `reusable` to write these fields once.  Example:  https://github.com/elastic/endpoint-package/blob/894313f80e3157c5d1f22363e15b5329d0345359/custom_schemas/custom_process.yml#L12-L16

In this case, I think it would be something like:
```yaml
  reusable:
    top_level: false
    expected:
      - { at: dll.Ext, as: device }
      - { at: file.Ext, as: device }
      - { at: process.Ext, as: device }
```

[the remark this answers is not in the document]
gabriellandau: Copy/paste (as is done here) is fine too

ferullo:
I hadn't realized there is a difference between mapping and indexing regarding storage size.

I'd like these fields to be mapped/defined somehow because we've gotten into a risky workflow where Endpoint keeps adding fields without going through a PR process in this repo by adding exceptions in Endpoint's repo to Endpoint testing in that would otherwise fail. The longer that list of exceptions gets the more it feels appropriate to add more exceptions, putting us in a downward spiral.

There's also the fact that these mappings define the [Endpoint data schema documentation](https://docs.elastic.co/integrations/endpoint) so it feels arbitrary to not have these things documented. At least a few ought to be documented (data in events/alerts). The info in Endpoint metrics documents could arguably remain undocumented but since a lot of that document is already documented it feels wrong to fail to document all of it.

Still, the ultimate size of document storage in Elasticsearch is of primary concern. My presumption was not indexing this data would address that size-bloat concern while allowing the documentation/testing/pr-process concerns to be met. I'll be curious what you find about mapping-but-not-indexing's effect on storage size. If size still winds up a concern we can figure something else out.

custom_schemas/custom_call_stack.yml
```diff
@@ -12,25 +12,44 @@
...
-        The DLL/module containing `instruction_pointer`.
+        The path to the DLL/module containing `instruction_pointer`.
 
     - name: instruction_pointer
       level: custom
       type: keyword
       description: >
         The return address of this stack frame.
 
+    - name: memory_section.memory_address
```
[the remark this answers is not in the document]
gabriellandau: Hrm.  `memory_section.memory_address` appears to have come first, a carry-over from Endgame schema.  It's used for memory protection call stacks alongside `memory_size`.

`memory_section.address` appears to be new to Elastic Credential Access events (8.3.0).  Can we remove it from the schema, and change CredAccess in 8.3.1 over to `memory_address`?

I'm not finding any references to `memory_section.size` - we may be okay to remove that.

custom_schemas/custom_dll.yml
```diff
@@ -109,6 +109,48 @@
         These defense evasions can make it harder to inspect a process and/or cause abnormal OS behavior.
         Examples tools that can cause defense evasions include KnownDlls hijacking and PPLDump.
 
+    - name: Ext.device.bus_type
+      level: custom
+      type: keyword
+      short: FILL ME IN
+      description: >
+        FILL ME IN
```
Trinity2019, line 117:
```suggestion
        Bus type of the device, such as Nvme, Usb, FileBackedVirtual,... etc.
```

custom_schemas/custom_dll.yml
```diff
@@ -109,6 +109,48 @@
         These defense evasions can make it harder to inspect a process and/or cause abnormal OS behavior.
         Examples tools that can cause defense evasions include KnownDlls hijacking and PPLDump.
 
+    - name: Ext.device.bus_type
+      level: custom
+      type: keyword
+      short: FILL ME IN
```
Trinity2019, line 115:
```suggestion
      short: Bus type of the device.
```

custom_schemas/custom_dll.yml
```diff
@@ -109,6 +109,48 @@
...
+      description: >
+        FILL ME IN
+
+    - name: Ext.device.dos_name
+      level: custom
+      type: keyword
+      short: FILL ME IN
+      description: >
+        FILL ME IN
```
Trinity2019, line 124:
```suggestion
        DOS name of the device.
        DOS device name is in the format of driver letters such as C:, D:,...
```

custom_schemas/custom_dll.yml
```diff
@@ -109,6 +109,48 @@
...
+
+    - name: Ext.device.dos_name
+      level: custom
+      type: keyword
+      short: FILL ME IN
```
Trinity2019, line 122:
```suggestion
      short: DOS name of the device.
```

custom_schemas/custom_dll.yml
```diff
@@ -109,6 +109,48 @@
...
+      description: >
+        FILL ME IN
+
+    - name: Ext.device.nt_name
+      level: custom
+      type: keyword
+      short: FILL ME IN
```
Trinity2019, line 129:
```suggestion
      short: NT name of the device.
```

custom_schemas/custom_dll.yml
```diff
@@ -109,6 +109,48 @@
...
+      description: >
+        FILL ME IN
```
Trinity2019, line 131:
```suggestion
        NT name of the device.
        NT device name is in the format such as:
        \Device\HarddiskVolume2
```

custom_schemas/custom_dll.yml
```diff
@@ -109,6 +109,48 @@
...
+
+    - name: Ext.device.product_id
+      level: custom
+      type: keyword
+      short: FILL ME IN
```
Trinity2019, line 136:
```suggestion
      short: ProductID of the device.
```

custom_schemas/custom_dll.yml
```diff
@@ -109,6 +109,48 @@
...
+      description: >
+        FILL ME IN
```
Trinity2019, line 138:
```suggestion
        ProductID of the device. It is provided by the vendor of the device if any.
```

custom_schemas/custom_dll.yml
```diff
@@ -109,6 +109,48 @@
...
+
+    - name: Ext.device.serial_number
+      level: custom
+      type: keyword
+      short: FILL ME IN
```
Trinity2019, line 143:
```suggestion
      short: Serial Number of the device.
```

custom_schemas/custom_dll.yml
```diff
@@ -109,6 +109,48 @@
...
+      description: >
+        FILL ME IN
```
Trinity2019, line 145:
```suggestion
        Serial Number of the device. It is provided by the vendor of the device if any.
```

custom_schemas/custom_dll.yml
```diff
@@ -109,6 +109,48 @@
...
+
+    - name: Ext.device.vendor_id
+      level: custom
+      type: keyword
+      short: FILL ME IN
```
Trinity2019, line 150:
```suggestion
      short: VendorID of the device.
```

custom_schemas/custom_dll.yml
```diff
@@ -109,6 +109,48 @@
...
+      description: >
+        FILL ME IN
```
Trinity2019, line 152:
```suggestion
        VendorID of the device. It is provided by the vendor of the device.
```

ferullo (commented):
Thanks @Trinity2019

custom_schemas/custom_file.yml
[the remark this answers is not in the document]
ferullo: I chose to copy/paste. It seemed like I'd have to create a new `custom_device.yml` file to make use of the reuse, right?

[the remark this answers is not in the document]
gabriellandau: Yep

custom_schemas/custom_call_stack.yml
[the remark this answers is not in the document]
ferullo: I see `call_stack.memory_region.[address|size]` in `custom_subsets/elastic_endpoint/alerts/malware_event.yaml`, `custom_subsets/elastic_endpoint/alerts/memory_protection_event.yaml`, and `custom_subsets/elastic_endpoint/alerts/ransomware_event.yaml`. Is there any reason I shouldn't remove `address` and `size` from all three and add `memory_address` and `memory_size`?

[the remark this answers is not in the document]
gabriellandau: > Is there any reason I shouldn't remove address and size from all three and add memory_address and memory_size?

From what I can tell, that would be a reasonable course of action.  I can do that in a follow-up PR if you don't want to do it here.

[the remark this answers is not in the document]
ferullo: Oh, based on git blame that field is really old. I wonder why it was added so long ago?

[Malware git blame](https://github.com/elastic/endpoint-package/blame/master/custom_subsets/elastic_endpoint/alerts/malware_event.yaml) and [PR that added it](https://github.com/elastic/endpoint-package/pull/23) on Jul 1, 2020

[Memory protection git blame](https://github.com/elastic/endpoint-package/blame/master/custom_subsets/elastic_endpoint/alerts/memory_protection_event.yaml) and [PR that added it](https://github.com/elastic/endpoint-package/pull/144) on Apr 15, 2021

[Ransomware git blame](https://github.com/elastic/endpoint-package/blame/master/custom_subsets/elastic_endpoint/alerts/ransomware_event.yaml) and [PR that added it](https://github.com/elastic/endpoint-package/pull/96) on Oct 26, 2020

It's unclear to me why it is Malware and Ransomware docs in the first place.

gabriellandau:
> @joe-desimone or @gabriellandau is it ok that the device.* fields won't be indexed or exceptionable?

Even without indexing, @Samirbous will still be able to write endpoint rules based on them.

I could imagine users wanting to hunt for processes run from USB drives (`event.category: process and process.Ext.device: Usb`), for example, but I defer to @Samirbous for how likely that is.

joe-desimone:
I think indexing them makes sense for users to write their own rules and for us to build insider threat type detections on this data in the stack. For example, in a UEBA/Entity analytics type use case, we have an interest in looking for users writing large volumes of data to USB drives.

custom_subsets/elastic_endpoint/alerts/memory_protection_event.yaml
```diff
@@ -442,11 +443,12 @@ fields:
...
                 fields:
                   module_path: {}
+                  module_name: {}
                   instruction_pointer: {}
                   memory_section:
                     fields:
-                      address: {}
-                      size: {}
+                      memory_address: {}
+                      memory_size: {}
```
gabriellandau, line 451: 💯

custom_subsets/elastic_endpoint/alerts/malware_event.yaml
```diff
@@ -594,8 +595,8 @@ fields:
                   instruction_pointer: {}
                   memory_section:
                     fields:
-                      address: {}
-                      size: {}
+                      memory_address: {}
```
gabriellandau, line 598: We don't need to do this now, but I'm pretty sure we can get rid of `thread` entirely from malware events.

gabriellandau (approved):

ferullo:
@kevinlog @pzl I think this is ready for approval and merge if it looks good (I'm not saying this doesn't need any more legit PR scrutiny, just that I think my major uncertainties have need discussed and decided upon).

I'm unsure of applicability of the unchecked items in this PR's description. I don't think any of those steps apply to this PR but defer to you to tell me if I'm wrong on that.

pzl (approved):
lets do it

ferullo:
Awesome. Am I good to merge? Notably, I want to make sure it's clear I haven't had a Kibana instance use these PR changes because I don't really have a dev set up to do that.

Issue #267, opened 2022-06-27
Adapt endpoint package to package storage v2


kevinlog:
Change the `endpoint` package to adhere to the [package storage v2](https://github.com/elastic/elastic-package/blob/main/docs/howto/use_package_storage_v2.md#use-package-storage-v2) rquirements and to upload signed packages.

AC:
- [ ] Update the job that pushes the Endpoint package to package-storage.

Notes:
- Reference this ticket for more info: https://github.com/elastic/ingest-dev/issues/1119

mtojek:
@kevinlog Could you please share a link/reference to your delivery pipeline? I may find some time to help you adjust it.

mtojek:
No worries about shipping. We will make sure that you're onboarded before migration.

pzl:
> prerelease based only on the semver.

FYI we currently use some illegal versioning. We haven't aligned our process to use the correct prerelease version strings to spec yet.

---


The pipeline sounds good. We have a Jenkinsfile and are set up with CI within our repo (and it is on the beats-ci cluster). It builds and checks on PRs/merge, etc. Happy to integrate or add the jobs in our repo's `.ci/` so it can be in sync with repo build changes? 
It would also be nice, if its simple while we're doing this, to publish pre-release versions on every commit (/merge) to our main branch. So we auto-deploy prereleases, but manual deploy job for release versions?

mtojek:
> It would also be nice, if its simple while we're doing this, to publish pre-release versions on every commit (/merge) to our main branch. So we auto-deploy prereleases, but manual deploy job for release versions?

Yes, we can go that way:
1. Publish a prerelease package on every commit.
2. Provide a pipeline parameter to publish a release (not prerelease).

mtojek:
@pzl 

These are the leftovers:
1. Remove the release script.
2. Update docs as there won't be promotions anymore.
3. Automate version bump (for development versions).
4. Enable full package validation, disable `legacy_package` mode.

Would you like me to create a new issue for those or do you prefer to keep this one?

PR #272, opened 2022-07-20
Publish endpoint-package to Package Storage v2

mtojek:
Issue: https://github.com/elastic/endpoint-package/issues/267

The PR modifies the existing Jenkins pipeline to automatically build and publish the endpoint-package to Package Storage v2. It won't be required to publish and promote files manually.

Steps:
1. Check if the endpoint-package hasn't been released yet, based on the manifest's `version` property. 
2. Sign the endpoint-package with the Infra signing pipeline.
3. Upload the package.zip and package.zip.sig (signature) to Package Storage v2.

Limitation:
1. Pipeline doesn't bump up the package version (not in this PR). Manual adjustment is needed.
2. Package is built using elastic-package, but it's still invalid.

Next steps:
1. Remove the release script.
3. Update docs as there won't be promotions anymore.
4. Automate version bump (for development versions).
5. Enable full package validation, disable `legacy_package` mode.


pzl:
> `bump2version` tool manually

We do not have any particular attachment to keep using this tool.

I am fine using another versioning scheme. Incrementing counter on commit-since-last-release seems perfectly fine. I don't think `-dev.N` is a valid versioning scheme according to the spec. I think we can also move to align whatever the proper prerelease semantics should be for the name.

Performing the actual move of prerelease-to-cutting a full `8.4.0` release can remain a manual job for myself and equivalents  for now. It's already quite manual anyway

mtojek:
It depends on what we consider as MVP :) Without the `commit-since-last-release` procedure, you would have to bump the `-dev.N` version manually. I'm wondering if I should split this task into 2 PRs to unblock migration to v2, and focus on publishing logic first.

Let me know what you think.

Makefile
```diff
@@ -60,9 +60,9 @@ endif
 SED := gsed
 endif
 
-
+# FIXME use `elastic-package check` to enable linting
 all: $(VENV_DIR) $(ECS_TAG_REF) $(PKG_FIELDS_TARGETS) $(DOC_TARGET) $(ESTC_PKG_BIN) $(SCHEMA_TARGETS)
-	cd $(PKG_DIR) && $(ESTC_PKG_BIN) format
+	cd $(PKG_DIR) && $(ESTC_PKG_BIN) format && $(ESTC_PKG_BIN) build -v --skip-validation
```
pzl, line 65: If I understand the log output here:

`elastic package build -v --skip-validation` is doing:

- obviously, skipping some checks and linting since we still have the old-style transforms which will fail it
- copying `./package/endpoint` to `./build/packages/endpoint/$(VERSION)`
- doing various package things I don't know
- zipping `./build/packages/endpoint/$(VERSION)` into `./build/packages/endpoint-$(VERSION).zip`


Where `$(VERSION)` I assume is pulled from `package/endpoint/manifest.yml`


Is that understanding correct?

---

The only comment I'd make here is that the general case `make` in our repo is used for devs to build our source field definitions (in custom_subsets and custom_schemas) into `package/endpoint/data_stream`. It will likely not be a part of the regular development cycle to then copy the package and make the zip.  (depending on how local development of packages will proceed. I believe the usual so far is to spin up a docker of `package-registry/distriution:production` and volume-mount the package dir to the correct place. Will that still work, or will zip be needed for local dev?)

If local development and installation of the package doesn't require the zip, then I might move the `elastic-package build` step to another target like `build-package`, and add that step to CI.

jsoriano (approved):

Makefile
mtojek: @pzl 

> Is that understanding correct?

Yes, it is. The steps that you don't know are: encoding dashboards, and importing fields from ECS ([source](https://github.com/elastic/elastic-package/blob/main/internal/builder/packages.go#L141)) 

> I believe the usual so far is to spin up a docker of package-registry/distriution:production and volume-mount the package dir to the correct place. Will that still work, or will zip be needed for local dev?)

Package Registry is able to recognize both distribution forms - zip or unpacked directory. We want to keep only .zip files, but the unpacked form is also stored in the `build` directory to overcome some technical challenges. We may change this behavior in the future to interact only with .zip files.

> If local development and installation of the package doesn't require the zip, then I might move the elastic-package build step to another target like build-package, and add that step to CI.

I don't want to break your development workflow, folks. Let me know if you want me to refactor the pipeline, but remember that I'd like to focus on the publishing part in this pull-request.

pzl (approved):

PR #273, opened 2022-07-20
add process.end

ferullo:
## Change Summary

Add `process.end` from ECS into Endpoint process documents.

### Sample values

See blow

Sample document:
Heavily redacted/snipped example:

```json
{
    "message": "Endpoint process event",
    "process": {
        "end": "2022-07-18T21:05:19.9419692Z"
    }
}
```

## Release Target

8.4

## Q/A

N/A

### For mapping changes:

### For Transform changes:


ferullo:
👍 done

package/endpoint/data_stream/process/sample_event.json
```diff
@@ -159,4 +160,4 @@
         "name": "SYSTEM",
         "id": "S-1-5-18"
     }
-}
\ No newline at end of file
+}
```
pzl, line 163: I vaguely recall the linter being very particular about not liking final newlines. Unfortunately I don't think I can relax its rules. We will see if CI complains

pzl: yep, 

```
[2022-07-20T16:39:57.255Z] + git diff-index --exit-code HEAD --
[2022-07-20T16:39:57.255Z] :100644 100644 177a6e95dde05a0e739866f032964d3127fa4ebd 0000000000000000000000000000000000000000 M	package/endpoint/data_stream/process/sample_event.json
script returned exit code 1
```

after a `format` is run, it seems to standardize no-ending-newline and then we detect a git diff on that.

For the future I will see if we can relax that, but for now the fix here is to delete trailing \n

ferullo: Thanks for the heads up. It should be fixed now.

Trinity2019 (approved):

PR #276, opened 2022-08-01
Volume device type

ferullo:
## Change Summary

This adds `volume_device_type`.

### Sample values

Sample document:

None

## Release Target

Ideally for 8.4

## Q/A

N/A

### For mapping changes:

- [x] I ran `make` after making the schema changes, and committed all changes
- [ ] If these field(s) are "exception"-able, I made a companion PR to Kibana adding it (see [Readme](https://github.com/elastic/endpoint-package#exceptionable))
- [ ] If this is a `metadata` change, I also updated both transform destination schemas to match

### For Transform changes:

- [ ] The new transform successfully starts in Kibana
- [ ] The corresponding transform destination schema was updated if necessary

PR #281, opened 2022-08-10
replace file.Ext.* with existing fields within alerts

pzl:
## Change Summary

in alerts, we had a glob-field (threat.enrichments or intel ... file.Ext.*) which pulls in any `file.Ext..` field as they are added in the future. This is probably pulling in more fields than intended, so we are taking the practice of replacing any `*` with explicit field trees as they come up.

So this is replacing the `...file.Ext.*` with whatever fields existed up to now (knowing that we need some further field pruning in the future). This should result in a net-nothing diff, aside from below:

The exception to this is `.. Ext.device.volume_device_type` which was just added in #276 and explicitly noted as an unwanted field. So those are excluded. We do see those being removed from the final product

## Release Target

8.4 stack (8.4.1 package)

### For mapping changes:

### DIFF
--- a/custom_subsets/elastic_endpoint/alerts/malware_event.yaml
+++ b/custom_subsets/elastic_endpoint/alerts/malware_event.yaml
@@ -186,7 +186,73 @@ fields:
                   type: {}
                   uid: {}
                   Ext:
-                    fields: "*"
+                    fields:
+                      code_signature:
+                        fields:
+                          exists: {}
+                          status: {}
+                          subject_name: {}
+                          trusted: {}
+                          valid: {}
+                      device:
+                        fields:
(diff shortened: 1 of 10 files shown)


kevinlog (approved):

paul-tavares (approved):
I'm 👍 , but: I think you should have someone from the Protection's team also review/approve this since I think they are the ones that consume these doc's in kibana as part of the detection engine.

pzl:
@paul-tavares usually, yes, good idea. This PR is pretty much a net-neutral, nothing-changing situation. The only change being, something that was accidentally added in a PR yesterday that we agreed to get in a followup (this) PR.

PR #270, opened 2022-07-13
Transform schema v2

pzl:
## Change Summary

Updated the transform definitions to match the new [Transform Spec](https://github.com/elastic/package-spec/pull/307)

this involved mostly renaming files and translating from json to yaml

I tested these package changes against this PR: https://github.com/elastic/kibana/pull/140046

You can see that the new transforms are installed below.

image

I added an Endpoint to make sure the data was still making it's way to the Endpoint list.

image

## Release Target

next

### For mapping changes:

- [x] I ran `make` after making the schema changes, and committed all changes
- [x] If this is a `metadata` change, I also updated both transform destination schemas to match

### For Transform changes:

- [x] The new transform successfully starts in Kibana
- [x] The corresponding transform destination schema was updated if necessary

Upgrade (from current json-based transform) & downgrade operate successfully

PR #295, opened 2022-08-31
add new `.fleet-agents` fields to united index

joeypoon:
## Change Summary

Updating united index mapping to include new fields in `.fleet-agents`: https://github.com/elastic/elasticsearch/pull/89599. Left both as `enabled: false` since I don't believe we'll need to be searching on them.

## Release Target

8.5

### For mapping changes:


pzl (approved):
lgtm.

note to self, add these fields in the changeover #270

PR #297, opened 2022-09-12
[Security Solution] Fix united index template

kevinlog:
## Change Summary
This PR fixes a bug in the `metadata-united` index pattern.  Before, Fleet would fail to install the Endpoint package citing the `metadata_united` index pattern as invalid.

Before we were using the `enabled` field like this:
```
"last_checkin_message": {
    "type": "text",
    "enabled": false
},
```

`"enabled"` does not look like it's compatible with `"type": "text"` .  I assumed the intention was to not index the field, so I added `"index:" "false"` to replace `"enabled"`.  After this change, Fleet is able to install the Endpoint package.

## Release Target

`8.5`

### For mapping changes:

### For Transform changes:

### DIFF
--- a/package/endpoint/elasticsearch/index_template/metrics-metadata-united.json
+++ b/package/endpoint/elasticsearch/index_template/metrics-metadata-united.json
@@ -336,11 +336,11 @@
                                 },
                                 "last_checkin_message": {
                                     "type": "text",
-                                    "enabled": false
+                                    "index": false
                                 },
                                 "last_updated": {
                                     "type": "date"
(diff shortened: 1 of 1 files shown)


paul-tavares (approved):
LGTM - but I don't really know much in this area. 

Dan/Joey should review

joeypoon (approved):
> "enabled" does not look like it's compatible with "type": "text"

ah, TIL. thanks for looking into this.

pzl (approved):

PR #325, opened 2022-11-29
Update 8.6 branch Jenkins file to publish 8.6.x line Endpoint packages

kevinlog:
## Change Summary

This updates the Jenkins file for the `8.6` branch to publish new packages on the `8.6.x` line.  It should publish any incremented `8.6.x` packages merged to the `8.6` branch.


pzl (approved):
this is one way to do it!

for our purposes today, and mostly going forward, this should be ok.

Only watch-out is if the Jenkinsfile ever gets refactored or changed on `main`, and we need to backport stuff. The small difference here may be missed and re-broken in the future. But the likelihood of all that happening is.. pretty low. Just wanted to note the possibility so we might remember if it _were_ to occur

PR #330, opened 2023-01-09
add examples of ransomware alerts

ayfaouzi:
## Change Summary

Adding examples of ransomware alerts upon request of a customer.


ferullo (commented):
Thanks @ayfaouzi . These look like they were pulled from Elasticsearch after pipeline enrichment occurred. Can you use the original documents Endpoint generates? These don't match the ECS schema.

ayfaouzi:
@ferullo PR feedback addressed, does it look fine now ?

pzl:
This seems to have stalled -- not sure if we still want this change? Just reopen if we want this reviewed and added

PR #337, opened 2023-01-30
Update unsupported u64 type to unsigned_long

kevinlog:
## Change Summary

When installing the new Endpoint package with the 8.7.0 changes, the addition of a type `u64` is causing the package to fail to install.  This changes this to be `unsigned_long` which is a supported type and can be installed.

The merge that introduced the unsupported type was only applied to the 8.7.0 line, so we just need to fix it here before releasing the new package.  The problem has not been released to production.

PR with the u64 type: https://github.com/elastic/endpoint-package/pull/329

Console error with u64 type:
image

new installed template with `unsigned_long`
image

### Sample values

Sample document:

```json


```

### For mapping changes:

### For Transform changes:

### DIFF
--- a/custom_schemas/custom_dll.yml
+++ b/custom_schemas/custom_dll.yml
@@ -128,7 +128,7 @@
         
     - name: Ext.size
       level: custom
-      type: u64
+      type: long
       short: Size of DLL
       description: >
         Size of DLL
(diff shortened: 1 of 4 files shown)


gabriellandau (approved):

gabriellandau (approved):

ashokaditya (approved):
🚀

PR #346, opened 2023-02-24
Update package overview

kevinlog:
## Change Summary

Update the package overview description to add more detail.

image

## Release Target

Kibana: 8.7.0
Package: 8.7.1

### For mapping changes:

~~- [ ] If these field(s) are "exception"-able, I made a companion PR to Kibana adding it (see [Readme](https://github.com/elastic/endpoint-package#exceptionable))~~
~~- [ ] If this is a `metadata` change, I also updated both transform destination schemas to match~~

### For Transform changes:

~~- [ ] The new transform successfully starts in Kibana~~
~~- [ ] The corresponding transform destination schema was updated if necessary~~

### DIFF
--- a/doc_templates/endpoint/docs/README.md
+++ b/doc_templates/endpoint/docs/README.md
@@ -1,6 +1,16 @@
 # Elastic Defend Integration
 
-This integration sets up templates and index patterns required for Elastic Defend.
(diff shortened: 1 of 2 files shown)


pzl (approved):
looks good

small note for release targeting is that this will be 8.7.1

PR #353, opened 2023-03-06
update transforms to be unattended

joeypoon:
## Change Summary

Update current + united metadata transforms to be `settings.unattended: true`.

Related: https://github.com/elastic/security-team/issues/6165

## Release Target

`8.8`

### For Transform changes:

### DIFF
--- a/package/endpoint/elasticsearch/transform/metadata_current/default.json
+++ b/package/endpoint/elasticsearch/transform/metadata_current/default.json
@@ -28,5 +28,8 @@
             "field": "event.ingested",
             "delay": "1s"
         }
+    },
+    "settings": {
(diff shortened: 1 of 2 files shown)


joeypoon:
> Does fleet respect this `settings` key? Is the transform installed with this setting?

yes and yes. this won't change the transform functionally. this just tells ES to manage the transform's health and to always retry if in failing state: https://www.elastic.co/guide/en/elasticsearch/reference/current/put-transform.html (sorry no anchors on page, have to search for `unattended`).

pzl (approved):
looks good.

I bet the linters are complaining about the final newline in the file. Should pass if you remove it

PR #354, opened 2023-03-15
Update Endpoint package categories

jen-huang:
## Change Summary

Update integration category per decision in elastic/ingest-dev#1349.

Related to https://github.com/elastic/integrations/pull/5123.

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -18,7 +18,7 @@ values =
 [bumpversion:file:./package/endpoint/manifest.yml]
 search = 
 	version: {current_version}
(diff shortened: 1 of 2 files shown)


joeypoon (changes requested):
I believe we need to also update two additional instances of `categories` in `.bumpversion.cfg`: https://github.com/elastic/endpoint-package/blob/fdd1ce6d5cb7719929f203ba51b15221bc754cd7/.bumpversion.cfg#L21

joeypoon (approved):
🙆‍♂️

PR #355, opened 2023-03-20
Add fields connected to rules and alerts

tomsonpl:
Closes: https://github.com/elastic/security-team/issues/5929

Not sure if we need rule_name, but from what I remember, it was agreed that we wanted it.

### DIFF
--- a/custom_schemas/custom_endpoint_actions.yml
+++ b/custom_schemas/custom_endpoint_actions.yml
@@ -67,6 +67,7 @@
 
     - name: data.rule_name
       type: keyword
+      index: false
       level: custom
       short: rule name
       description: >
--- a/custom_schemas/custom_endpoint_actions.yml
+++ b/custom_schemas/custom_endpoint_actions.yml
@@ -58,21 +58,6 @@
       description: >
         A comment that describes the action that is requested
 
-    - name: data.rule_id
-      type: keyword
-      level: custom
-      short: rule id
-      description: >
-        ID of the rule that triggered the action
-
-    - name: data.rule_name
-      type: keyword
-      index: false
-      level: custom
-      short: rule name
-      description: >
-        Name of the rule that triggered the action
-
     - name: type
       type: keyword
       level: custom
--- a/custom_schemas/custom_endpoint_actions.yml
+++ b/custom_schemas/custom_endpoint_actions.yml
@@ -36,10 +36,10 @@
       description: >
         The action request information
 
-    - name: data.alert_ids
+    - name: data.alert_id
       type: keyword
       level: custom
-      short: alert ids
+      short: alert id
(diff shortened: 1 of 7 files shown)


tomsonpl:
@pzl any idea if `rule_name` should be a `keyword` or just `text` ?

pzl:
> any idea if `rule_name` should be a `keyword` or just `text` ?

are we ever going to filter/search via name? I would imagine in filtering scenarios we probably know the rule_id and will be exact matching on that, right? 

If for some reason we will filter on the name, do we know the full string to match (e.g. find rule by name "Unusual Network Activity")? then keyword. If we expect to filter on partial word matches (e.g. match the previous rule name based on the search string "network") then text could be ok.

If we don't need to filter by name, then I would set it to keyword, and `index: false`. Or just don't map it at all

tomsonpl:
@pzl @dasansol92 could you take a look please? Thank you!

pzl (approved):
looks okay, I think things are added in the right place. Added @ashokaditya to review the intent of whats being added

custom_schemas/custom_endpoint_actions.yml
```diff
@@ -35,6 +35,13 @@
       short: data
       description: >
         The action request information
+
+    - name: data.alert_ids
+      type: keyword
+      level: custom
+      short: alert ids
+      description: >
+        List of alert ids that triggered the action
```
ashokaditya, line 44: You'd also need to edit the `custom_subsets/elastic_endpoint/actions/actions.yaml` file. If you want the fields to also exist on the response doc then you need to edit the `custom_subsets/elastic_endpoint/action_responses/action_responses.yaml`.

ashokaditya (changes requested):
I added some suggestions to update the mappings. Let me know if you need help or have questions.

ashokaditya (commented):
Took another pass at reviewing. I have a question and suggestions.

custom_schemas/custom_endpoint_actions.yml
```diff
@@ -35,6 +35,13 @@
       short: data
       description: >
         The action request information
+
+    - name: data.alert_id
+      type: keyword
+      level: custom
+      short: alert id
+      description: >
+        List of alert ids that triggered the action
```
ashokaditya, line 44: Thanks for the changes. The fact that this is a `keyword` we can store an array of strings here s, so you can have multiple alert ids in the doc. Just that we should conform to ECS naming best practices.

ashokaditya (approved):

ashokaditya (approved):

tomsonpl:
@ashokaditya Thank you for your feedback and bearing with me fixing it so many times :P

PR #359, opened 2023-04-03
Revert "update transforms to be unattended (#353)"

joeypoon:
## Change Summary

Revert `unattended` transform option. This change is causing package install to occasionally fail. Since `unattended` means the transform will never throw, fleet doesn't reattempt to reinstall the transform destination indices during transient package install failures leaving the destination indices uncreated.


ashokaditya (approved):

kevinlog:
checked this out and tried it.  Transforms are being installed correctly and I can use the generator to stream data in properly

image

image

PR #368, opened 2023-04-26
set up 8.9 dev cycle

pzl:
## Change Summary

Setup the 8.9 development cycle

`-next` convention is now reserved for use in the `changelog` file, as its intention is to relax the linter:

to allow the top-line `changelog` version to differ from the `manifest` version during development, where the `manifest` file version will likely change with each prerelease. Now the changelog version does not have to constantly change to match.

However, the manifest version _does_ need to continually change to actually perform releases. We cannot overwrite a previous release (even prerelease) with the same version name. So we are going back to the `-dev.N` convention for now, incrementing periodically. This will likely be automated in the future

### DIFF
--- a/package/endpoint/changelog.yml
+++ b/package/endpoint/changelog.yml
@@ -1,3 +1,8 @@
+- version: "8.9.0-next"
+  changes:
+    - description: TBD
+      type: enhancement
+      link: https://github.com/elastic/endpoint-package/pull/364
 - version: "8.8.0"
   changes:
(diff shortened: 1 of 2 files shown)


pzl:
CI is just failing because 8.9 stack is not ready for pulling yet. This can merge without that (although that means all incoming PRs will fail until 8.9 stack container exists)

ashokaditya (approved):

PR #370, opened 2023-05-04
ETW Threat-Intelligence API Event metrics

jdu2600:
## Change Summary

Adds new metrics fields for the API Event type and the ETW Threat-Intelligence API Event provider.

```
Endpoint.metrics.documents_volume.api_events.sent_bytes
Endpoint.metrics.documents_volume.api_events.sent_count
Endpoint.metrics.documents_volume.api_events.suppressed_bytes
Endpoint.metrics.documents_volume.api_events.suppressed_count
Endpoint.metrics.documents_volume.api_events.sources[].source
Endpoint.metrics.documents_volume.api_events.sources[].sent_bytes
Endpoint.metrics.documents_volume.api_events.sources[].sent_count
Endpoint.metrics.documents_volume.api_events.sources[].suppressed_bytes
Endpoint.metrics.documents_volume.api_events.sources[].suppressed_count
Endpoint.metrics.system_impact.threat_intelligence_events.week_idle_ms
Endpoint.metrics.system_impact.threat_intelligence_events.week_ms
```

## Release Target

These events are diagnostic in 8.8 and will be released in a future version.

# For mapping changes:

### DIFF
--- a/custom_schemas/custom_endpoint.yml
+++ b/custom_schemas/custom_endpoint.yml
@@ -596,6 +596,56 @@
       type: long
       description: Total size of suppressed documents
 
+    - name: metrics.documents_volume.api_events.sent_count
+      level: custom
+      type: long
+      description: Number of sent API Event documents
+
+    - name: metrics.documents_volume.api_events.sent_bytes
+      level: custom
+      type: long
+      description: Total size of API Event sent documents
+
+    - name: metrics.documents_volume.api_events.suppressed_count
+      level: custom
+      type: long
+      description: Number of suppressed API Event documents
+
+    - name: metrics.documents_volume.api_events.suppressed_bytes
+      level: custom
+      type: long
+      description: Total size of suppressed API Event documents
+
+    - name: metrics.documents_volume.api_events.sources
+      level: custom
+      type: object
+      description: An array of API Event document statistics per source
+
+    - name: metrics.documents_volume.api_events.sources.source
+      level: custom
+      type: string
+      description: API Event document source name
+
+    - name: metrics.documents_volume.api_events.sources.sent_count
+      level: custom
+      type: long
+      description: Number of sent API Event documents from source
+
+    - name: metrics.documents_volume.api_events.sources.sent_bytes
+      level: custom
+      type: long
+      description: Total size of API Event sent documents from source
+
+    - name: metrics.documents_volume.api_events.sources.suppressed_count
+      level: custom
+      type: long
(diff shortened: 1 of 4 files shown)


intxgo (approved):
LGTM

custom_schemas/custom_endpoint.yml
```diff
@@ -846,17 +896,17 @@
       index: false
       description: The total milliseconds spent queueing process events for the process over the last week
 
-    - name: metrics.system_impact.etw_events.week_ms
```
kevinlog, line 849: Can you add new mappings for these instead of renaming?  The reason being is that there will still be old Endpoints that stream data to these fields.

custom_schemas/custom_endpoint.yml
```diff
@@ -846,17 +896,17 @@
...
+    - name: metrics.system_impact.threat_intelligence_events.week_ms
       level: custom
       type: unsigned_long
       index: false
-      description: The total milliseconds spent on ETW-based events for the process over the last week
+      description: The total milliseconds spent on ETW Threat-Intelligence events for the process over the last week
 
-    - name: metrics.system_impact.etw_events.week_idle_ms
```
kevinlog, line 855: Same here, can you add a new mapping instead of renaming the old one?  We would want to still be backwards compatible with old Endpoints

custom_schemas/custom_endpoint.yml
jdu2600: Ah. Sorry for not providing the background on this. I've been removing vestigial code in `endpoint-dev` - and just did the same here automatically.

So Endpoint has never generated `etw_events` metrics. This appears to have been added optimistically because of a feature in endgame and never used.
cc: @nfritts

kevinlog: thanks for the explanation.  This should be fine to merge as long as we're sure there's no backwards compatibility issues with older Endpoints

custom_schemas/custom_endpoint.yml
```diff
@@ -596,6 +596,56 @@
...
+      description: Total size of suppressed API Event documents
+
+    - name: metrics.documents_volume.api_events.sources
+      level: custom
+      type: object
+      description: An array of API Event document statistics per source
+
+    - name: metrics.documents_volume.api_events.sources.source
+      level: custom
+      type: string
```
jdu2600, line 626: This should have been `keyword`.
See https://github.com/elastic/endpoint-package/pull/373

PR #373, opened 2023-06-12
Fix mapping error by replacing string with keyword

kevinlog:
## Change Summary

Fixes a mapping error due to the inclusion of `type: string` in a recent change.  The proposed change is to use `type: keyword` in place.

Failing test where this was caught: https://github.com/elastic/kibana/issues/147640

## Release Target

8.9

### For mapping changes:


ashokaditya (approved):

dasansol92 (approved):

kevinlog:
Going to merge this as it's blocking downstream work, we can update the `type` again in a followup

PR #387, opened 2023-06-27
Setup Jenkins to publish from 8.9 branch

kevinlog:
## Change Summary

Setup Jenkins to publish from 8.9 branch


pzl (approved):
we may want to consider removing this check, or checking `main || ^[78]\.\d+$` as a followup

PR #389, opened 2023-07-17
add endpoint custom documentation to generate-docs

ferullo:
## Change Summary
This is a replacement PR for https://github.com/elastic/endpoint-package/pull/334

This PR is the first of four chained PRs. I've broken them up to make them reviewable. The final goal is to document what fields are actually available in each Endpoint document type.

1. (This PR) Add the code to read in new YAML documentation and render it in Markdown.
2. Add minimal YAML documentation that defines all the document types https://github.com/elastic/endpoint-package/pull/390
3. Update all YAML documentation to https://github.com/elastic/endpoint-package/pull/391
4. Render the YAML into pre-committed Markdown files https://github.com/elastic/endpoint-package/pull/392

Please at least skim all four PRs because seeing them all will help understand this PR better. Notably, I have comments asking for feedback in the second PR that are relevant to this PR.

## Release Target
**This will ideally be backported into 8.9**

PR #390, opened 2023-07-17
add minimal custom_documentation files

ferullo:
## Change Summary
This PR is the second of four chained PRs. I've broken them up to make them reviewable. The final goal is to document what fields are actually available in each Endpoint document type.

1. Add the code to read in new YAML documentation and render it in Markdown. (https://github.com/elastic/endpoint-package/pull/389)
2. (**This PR**) Add minimal YAML documentation that defines all the document types
3. Update all YAML documentation to https://github.com/elastic/endpoint-package/pull/391
4. Render the YAML into pre-committed Markdown files https://github.com/elastic/endpoint-package/pull/392

Please at least skim all four PRs because seeing them all will help understand this PR better.

## Release Target
**This will ideally be backported into 8.9**


custom_documentation/endpoint/data_stream/alerts/linux/linux_malicious_behavior_alert.yaml
```diff
@@ -0,0 +1,18 @@
...
+  data_stream: logs-endpoint.alerts-*
+  filter:
+    event.code: behavior
+    event.dataset: endpoint.alerts
+    event.module: endpoint
+    host.os.type: linux
+fields:
+  endpoint:
+  - '@timestamp'
+  details:
```
ferullo, line 16: see https://github.com/elastic/endpoint-package/pull/391/files#r1265607445

pzl (approved):

pzl (approved):

custom_documentation/endpoint/data_stream/alerts/linux/linux_malicious_behavior_alert.yaml
```diff
@@ -0,0 +1,21 @@
...
+    This alert is generated when a Malicious Behavior alert occurs.
+identification:
+  os:
+  - linux
+  data_stream: logs-endpoint.alerts-*
+  filter:
+    event.code: behavior
+    event.dataset: endpoint.alerts
+    event.module: endpoint
+    host.os.type: linux
```
jdu2600, line 13: Hmm. It feels odd that the filter includes none of the five core ECS categorization fields.

jdu2600:
```suggestion
    event.action: rule_detection
    event.kind: alert
    event.module: endpoint
    host.os.type: linux
```

jdu2600 (approved):
So this aptly describes the current state of our endpoint documents.

However, I would flag a couple of potential emergent inconsistencies.
We appear to have introduced a new field, `event.code`, rather than using the existing ECS `event.action` field for specific categorization beyond `event.type`.
On posix platforms, we are overloading the process (lifecycle) category with process activities that don't fit into other categories.

custom_documentation/endpoint/data_stream/file/linux/linux_file_delete.yaml
```diff
@@ -0,0 +1,16 @@
+overview:
+  name: Linux File Delete
+  description: >
+    This event is generated when a file is deleted.
+identification:
+  filter:
+    event.action: deletion
```
jdu2600, line 7: Is there any performance benefit to filtering on the standard categorization fields that have fixed values?
```suggestion
    event.type: deletion
```

ferullo:
@jdu2600 I think you've hit the nail on the head with "So this aptly describes the current state of our endpoint documents.".

For better or worse, right now the primary ECS categorization fields can't be used to break up Endpoint documents in a way that accomplishes the task of "documenting the fields in event types Endpoint generates". I assume because ECS limits the permitted values in the ECS categorization fields, Endpoint has leaned into using `event.action` (usually) to most directly describe what "event type" an event is. As an aside, that's probably necessary since, as an example, I doubt ECS would want to define "extended attribute delete" as a permitted value in any of those primary ECS categorization fields, yet Endpoint needs a way to distinguish such an event on macOS. But, again, as you say, whether this is good or bad or inevitable or not, it is where we are right now.

All these documentation files are just a filter and a corresponding list of fields that can occur in events/documents matching that filter, with a rule that every event/document matches one and only one filter. Although not strictly needed, as much as possible, I've also tried to be consistent in how filters are written (and within an index, filters are 100% consistent). So, I've used `event.action` as consistently as possible because that's what Endpoint uses consistently.

You're right that alerts are a frustrating oddity. The reason is that Endpoint uses `event.code` to distinguish the alert type, not `event.action`. While malicious behavior alerts do use `event.action=rule_detection` other Endpoint alerts set `event.action` to the system action that triggered the alert (for malicious behavior alerts there could be many actions that led to the rule triggering, which I assume is why the action is set to something more generic). In other words, `event.action` in alerts is primarily used to describe the system action that caused the alert, not to describe the detection that generated the alert.

Anyway, this documentation will be beta at first. If users don't like the way I've broken things up we can always change it. Or we can always break things up using a completely different, second set of filters.

intxgo (approved):

PR #398, opened 2023-07-31
Add `code_signature` mappings for API events

ferullo:
## Change Summary

This adds mappings for `code_signature` fields in API events. If there's no stability concern, it should be backported to 8.9.

### Sample values

See the sample document in the PR. `code_signature` is a standard field.

## Release Target

8.9 backport, ideally.

### For mapping changes:

### For Transform changes:


ferullo:
I tested this functionally and it works.

joeypoon (approved):

PR #399, opened 2023-08-03
add enabled_features to the heartbeat document

ferullo:
## Change Summary

Adds `Endpoint.state.enabled_features` to the heartbeat document. I removed the `custom_documentation` since these are moving to a `.logs` data stream.

### Sample values

See PR content.

## Release Target

8.10

### For mapping changes:


package/endpoint/data_stream/heartbeat/sample_event.json
```diff
@@ -11,5 +11,17 @@
     "message": "Endpoint heartbeat",
     "event": {
         "ingested": "2023-07-18T20:40:09.279939Z"
+    },
+    "Endpoint": {
+        "state": {
+            "enabled_features": [
```
ferullo, line 17: @joeypoon what do you think of this exact location? `Endpoint.state.host_isolation=true|false` already exists. It felt like this belonged along side it conceptually.

pzl (approved):
looks good to me, assuming the removal from `custom_documentation` was intentional

package/endpoint/data_stream/heartbeat/sample_event.json
joeypoon: make sense to me 👍

joeypoon (approved):

ferullo:
I tested this and it works with upcoming Endpoint changes to add this data.

PR #408, opened 2023-08-14
[backport][8.10] Revert unattended transforms (#407)

pzl:
## Change Summary

backport #407 to 8.10


kevinlog (approved):

kevinlog:
In order to actually publish, I think we need to do something like this: https://github.com/elastic/endpoint-package/pull/387

paul-tavares (approved):

PR #425, opened 2023-09-14
Add buildkite Build step

gogochan:
## Change Summary

This PR adds a dev pipeline that does following:
- [x] Buildkite build step - handling build and git diff test
- [x] Buildkite check step - handling static and pipeline test
- [x] Adding make target `elastic-package` to use in the pipeline.
- [x] Configure Buildkite to cancel intermediate build

Here's a run of this pipeline: https://buildkite.com/elastic/endpoint-package/builds/36

Github status does not block any PR yet. This can be enabled later.

I am leaning toward creating a separated PR for publishing `main` or other releasable branches, and will not integrate this using pr-bot.

image

### Sample values

Sample document:

```json


```

### For mapping changes:

### For Transform changes:

### DIFF
--- a/.buildkite/pipeline.yml.py
+++ b/.buildkite/pipeline.yml.py
@@ -35,15 +35,14 @@ def main():
             ],
         },
         {
-            "label": "Run static test",
-            "depends_on": "build",
-            "key": "static_test",
+            "label": "Run Static and Pipeline check",
+            "key": "check",
             "command": ".buildkite/scripts/check.sh",
             "agents": test_agent,
             "notify": [
                 {
                     "github_commit_status": {
-                        "context": "Buildkite Check Static Test",
+                        "context": "Buildkite Check",
                     },
                 },
             ],
@@ -52,8 +51,7 @@ def main():
             "label": "Publish",
             "command": "echo Do publish when it is ready",
             "depends_on": [
-                "static_test",
-                "pipeline_test",
+                "check",
             ],
             "notify": [
                 {
--- a/.buildkite/pipeline.yml.py
+++ b/.buildkite/pipeline.yml.py
@@ -2,6 +2,10 @@
 
 import json
 
+build_agent = {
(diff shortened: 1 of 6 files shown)


.buildkite/pipeline.yml.py
```diff
@@ -1,14 +1,83 @@
...
-        ]
-    }))
+    steps = [
+        {
+            "label": "Build",
+            "command": "make",
+            "key": "build",
+            #"agents": test_agent,
+            "artifact_paths": [
+                "build/packages/*.zip",
```
pzl, line 20: does this do anything? What's the side-effect of marking something an `artifact_paths`?

.buildkite/pipeline.yml.py
```diff
@@ -1,14 +1,83 @@
...
+            #"agents": test_agent,
+            "notify": [
+                {
+                    "github_commit_status": {
+                        "context": "Check Pipeline Test",
+                    },
+                },
+            ],
+        },
+        {
```
pzl, line 58: These will both require that a stack is running.  And the environment variables are set to point to whatever stack is use

pzl (commented):
Will buildkite execute this change as part of its run?

gogochan:
This is not complete yet, I'm pushing it up so that I can see validate the workflow.
Do not review at the moment.

.buildkite/pipeline.yml.py
gogochan: if there's anything under build/packages/*.zip, it stashes it. This behavior is similar to https://github.com/elastic/endpoint-package/blob/main/.ci/Jenkinsfile#L63C1-L64C1

gogochan:
This is now ready for a review.

pzl (approved):
lets goooooo

dwhyrock (approved):

gogochan:
https://github.com/elastic/security-team/issues/6786
for tracking purpose

PR #436, opened 2023-10-03
global artifact update age

intxgo:
## Change Summary

Add artifact up-to-date age indicator. It's equal to number of days since the local artifacts copy at Endpoint side was last time deemed up-to-date (the date when either Endpoint updated the artifacts or verified that it's artifacts are up-to-date with Elastic CDN).

Note, when artifacts are pinned to a specific snapshot date, the age always equal to the numbers of days since that date. In addition the initial artifacts shipped with the installation package age is equal to the number of days since the package creation date (endpoint build date).

### Sample values

Artifacts made or verified to be up-to-date today:
`"Endpoint.policy.applied.artifacts.global.update_age": 0`
Artifacts 50 days old
`"Endpoint.policy.applied.artifacts.global.update_age": 50`

Sample document:

Endpoint altert and Endpoint policy response:

```json
{
    "Endpoint": {
        "policy": {
            "applied": {
                "artifacts": {
                    "global": {
                        "update_age": 0,
                    },
}

```

### For mapping changes:

### For Transform changes:

### DIFF
--- a/custom_schemas/custom_endpoint.yml
+++ b/custom_schemas/custom_endpoint.yml
@@ -336,6 +336,11 @@
       type: object
       description: information about global protection artifacts applied.
 
+    - name: policy.applied.artifacts.global.update_age
+      level: custom
+      type: unsigned_long
+      description: number of days since global artifacts were made up-to-date
+
     - name: policy.applied.artifacts.global.version
       level: custom
       type: keyword


intxgo:
anyone can let me know if we still have to run the `make` locally` to add any generated files? I think it was deprecated

pzl:
@intxgo  `make` is still required. It catches a lot of misunderstandings between source files changed and what the output effect on the package definition is to move to a CI/build step instead of committed outputs.

Also please add the field to the sample json documents that are used for automated testing:

`package/endpoint/data_stream/<index name>/sample_event.json`

it sounds like this will end up in a few of them

ferullo:
This PR is targeted to `8.11.` Should it go to `main` and get a backport label instead?

intxgo:
> This PR is targeted to `8.11.` Should it go to `main` and get a backport label instead?

I don't know, can anyone confirm which branch should it go to. We want to add it to 8.11 release.

pzl:
Yes, 8.11.0 was released, so this will merge to `main` and backport to `8.11` branch, and we can make a `8.11.1` release

PR #435, opened 2023-10-03
additional process callstack custom_documentation fields

jdu2600:
## Change Summary

Add missing `process.parent.thread.Ext.call_stack` fields to custom_documentation.
These optional fields have existed since 8.8.0

PR #431, opened 2023-09-26
Add linux capabilities to process events

nicholasberlin:
## Change Summary

Add process.thread.capabiliites.permitted and process.thread.capabilities.effective fields to 8.11 custom documentation for Endpoint

## Release Target

8.11.0

PR #437, opened 2023-10-03
Backport `custom_documentation` #435 to 8.10.3 

gabriellandau:
Backport https://github.com/elastic/endpoint-package/pull/435 to 8.10.3.

### DIFF
--- a/custom_documentation/endpoint/data_stream/process/windows/windows_process_create.yaml
+++ b/custom_documentation/endpoint/data_stream/process/windows/windows_process_create.yaml
@@ -112,6 +112,9 @@ fields:
   - process.parent.name
   - process.parent.pid
   - process.parent.thread.Ext.call_stack.symbol_info
+  - process.parent.thread.Ext.call_stack.protection
+  - process.parent.thread.Ext.call_stack.callsite_leading_bytes
+  - process.parent.thread.Ext.call_stack.callsite_trailing_bytes
   - process.parent.thread.Ext.call_stack_contains_unbacked
   - process.parent.thread.Ext.call_stack_summary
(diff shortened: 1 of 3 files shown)


joeypoon (approved):

pzl:
I'm not entirely sure what the.. deployment or viewing/consumption process of the `custom_documentation` stuff is.

@ferullo is the stuff in `custom_documentation` versioned at all? It just points to this repo, right? Is it branch-aware?  If its not, then backport was probably unnecessary

ferullo:
It's not branch aware itself. But the issue is automated Endpoint testing validates all Endpoint documents against `custom_documentation` _and_ the package schema using a single submodule of this repo. The issue Gabe identified was that `main` [increased](https://github.com/elastic/endpoint-package/pull/431) the ECS version from 8.7 to 8.10.

Endpoint's 8.10 branch shouldn't be tested against that change, since that change is in endpoint-package 8.11, not endpoint-package 8.10. So, at least right now, we need a single commit hash of endpoint-package that has the 8.10 package schema in it and any `custom_documentation` changes that we also needed to make.

We could also just have made an `8.10-endpoint-testing` branch that off `8.10` that has the `custom_documentation` stuff in it. But since `custom_documentation` doesn't really affect the  package (especially in 8.11+ where it'll entirely be in the `custom_documentation/` directory) that didn't feel like worth the effort. I don't mind if you think it is, we can do that in the future.

This isn't a great long term solution. IMO, conceptually `custom_documentation` belongs in this repo, but functionally it's not working well. I can look into moving it out, but I'm not sure we're better off doing that as a part of a larger changes to move this package into the main integration repo?

PR #442, opened 2023-10-10
Replace more dotted keys

pzl:
## Change Summary

replace more instances of yaml dotted-key format to multiline format (for package-spec 3.0)

the semantics of this PR are intended to be identical, only formatting changes


joeypoon (approved):

package/endpoint/data_stream/action_responses/_dev/test/pipeline/test-common-config.yml
```diff
@@ -1,2 +1,3 @@
 dynamic_fields:
-  event.ingested: ".*"
```
andrewkroh, line 2: I think `dynamic_fields` is expecting a flat list of field names. Have you tried quoting the key name like `"event.ingested"`?

PR #443, opened 2023-10-10
object_type: keyword

pzl:
## Change Summary

for any field marked `type: object`, it must have the accompanying setting `object_type` per [package-spec](https://github.com/elastic/package-spec/pull/628). 

This sets all of our objects to `keyword`. 

### Sample values
N/A

Sample document:

N/A

## Release Target

8.12

### For mapping changes:

- [x] If these field(s) are "exception"-able, I made a companion PR to Kibana adding it (see [Readme](https://github.com/elastic/endpoint-package#exceptionable))
- [] If this is a `metadata` change, I also updated both transform destination schemas to match 
^^ ????

### For Transform changes:

- ~~[ ] The new transform successfully starts in Kibana~~
- ~~[ ] The corresponding transform destination schema was updated if necessary~~

### DIFF
--- a/custom_schemas/custom_call_stack.yml
+++ b/custom_schemas/custom_call_stack.yml
@@ -4,6 +4,7 @@
   group: 3
   short: Fields describing a stack frame.
   type: object
+  object_type: keyword
   description: >
     Fields describing a stack frame.  call_stack is expected to be an array where each array element represents a stack frame.
   reusable:
--- a/custom_schemas/custom_api.yml
+++ b/custom_schemas/custom_api.yml
@@ -4,6 +4,7 @@
   group: 2
   short: Fields describing an API call.
   type: object
+  object_type: keyword
   description: >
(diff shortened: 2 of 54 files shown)


joeypoon (approved):

custom_schemas/custom_call_stack.yml
```diff
@@ -4,6 +4,7 @@
   group: 3
   short: Fields describing a stack frame.
   type: object
+  object_type: keyword
```
andrewkroh, line 7: This will create a dynamic mapping for `call_stack.*` that maps any matching properties a `keyword`. If you are not expecting properties beyond what is already explicitly mapped, then I would recommend setting `type: group` and eliminating `object_type`.

I think all of the cases here where the a non-leaf field was changed to a dynamic mapping should be changed to `group`. Then the resulting index template would not change, and the package would be following package-spec.

PR #427, opened 2023-09-22
ETW Threat-Intelligence API events

jdu2600:
## Change Summary

This PR adds multiple new API event variants, including new call_stack fields.

### Sample values for new fields

```json
"api": {
    "behaviors": [
        "cross-process"
    ],
    "metadata": {
        "target_address_name": "Unbacked"
    },
    "name": "VirtualAllocEx",
    "parameters": {
        "address": 2904525045760,
        "allocation_type": "COMMIT|RESERVE",
        "protection": "R-X",
        "size": 16
    },
    "summary": "VirtualAllocEx( charmap.exe, NULL, 0x10, COMMIT|RESERVE, R-X )"
}
```
```json
"api": {
    "behaviors": [
        "hook_api",
        "cross-process"
    ],
    "metadata": {
        "target_address_name": "ntdll.dll",
        "target_address_path": "c:\\windows\\system32\\ntdll.dll"
    },
    "name": "VirtualProtectEx",
    "parameters": {
        "address": 140706538901504,
        "protection": "RWX",
        "protection_old": "R-X",
        "size": 17
    },
    "summary": "VirtualProtectEx( charmap.exe, ntdll.dll, 0x11, RWX, R-X )"
}
```
```json
"api": {
    "behaviors": [
        "cross-process"
    ],
    "metadata": {
        "target_address_name": "Unbacked"
    },
    "name": "MapViewOfFile2",
    "parameters": {
        "address": 2904641306624,
        "allocation_type": "0x0",
        "protection": "R-X",
        "size": 18
    },
    "summary": "MapViewOfFile2( charmap.exe, 0x12, R-X )"
}
```
```json
"api": {
    "behaviors": [
        "hook_api"
    ],
    "metadata": {
        "target_address_name": "amsi.dll",
        "target_address_path": "c:\\windows\\system32\\amsi.dll"
    },
    "name": "WriteProcessMemory",
    "parameters": {
        "address": 140706238576736,
        "size": 1
    },
    "summary": "WriteProcessMemory( Self, amsi.dll!AmsiScanBuffer, 0x1 )"
}
```
```json
"api": {
    "behaviors": [
        "execute_shellcode",
        "hardware_breakpoint_set",
        "cross-process"
    ],
    "metadata": {
        "target_address_name": "Unbacked"
    },
    "name": "SetThreadContext",
    "parameters": {
        "context_flags": 1048607,
        "r8": 8,
        "r9": 9,
        "rax": 10,
        "rbp": 3,
        "rbx": 11,
[...]

PR #445, opened 2023-10-26
Add memory_region to api events

jdu2600:
Add missing `[Target.]process.Ext.memory_region` fields to api events.
These optional fields were introduced in 8.11.0

### DIFF
--- a/custom_subsets/elastic_endpoint/api/api.yaml
+++ b/custom_subsets/elastic_endpoint/api/api.yaml
@@ -61,6 +61,8 @@ fields:
           entity_id: {}
           Ext:
             fields:
+              memory_region:
+                fields: "*"
               token:
                 fields:
                   integrity_level_name: {}
@@ -86,6 +88,8 @@ fields:
                 fields: "*"
               parameters:
                 fields: "*"
+          memory_region:
+            fields: "*"
           token:
             fields:
               integrity_level_name: {}
--- a/package/endpoint/data_stream/api/fields/fields.yml
+++ b/package/endpoint/data_stream/api/fields/fields.yml
@@ -36,6 +36,718 @@
       object_type: keyword
       description: Object for all custom defined fields to live in.
       default_field: false
+    - name: process.Ext.memory_region.allocation_base
+      level: custom
+      type: unsigned_long
+      description: Base address of the memory allocation containing the memory region.
+      example: 2431737462784
+      default_field: false
+    - name: process.Ext.memory_region.allocation_protection
+      level: custom
+      type: keyword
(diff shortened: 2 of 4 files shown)


jdu2600:
🤔 Why do the generated files from my build have different whitespace to the existing files?

gabriellandau (commented):
Code LGTM but red build.

jdu2600:
> Code LGTM but red build.

The five `Required` items are green - which is sufficient I believe.

The failures look like the same `KeyError: 'kibana.version'` as [last month](https://github.com/elastic/endpoint-package/pull/435#issuecomment-1745416358) - which is part of the buildkite migration.

gabriellandau (approved):

gabriellandau:
I'm not going to block a merge over this, but could you follow up with a PR that adds the new fields to the [sample API event](https://github.com/elastic/endpoint-package/blob/c861c833f7eaa8099c7acbd18e87360b9caeec6f/package/endpoint/data_stream/api/sample_event.json#L55) so their inclusion in enforced by `make static-test`?

jdu2600:
> `api/sample_event.json`

That's a credential_access API event sample - which I didn't want to overwrite. 
I looked at that in my last PR, but couldn't see a way to add samples for all of the new API event variants.

gabriellandau:
> That's a credential_access API event sample - which I didn't want to overwrite.

They don't have to be real events with real meaning.  The point is just to exercise the schema validation code.

ferullo:
Should this be targeting 8.11 or main? Would we really release an updated endpoint package for 8.11?

gabriellandau:
@jdu2600 the data's already there but not in the mapping, right?  If so, that's arguably a bug.

pzl:
Echoing @gabriellandau, I would love to see values for these fields added to the sample file.

The `sample_event.json` file does not need to semantically make any sense. It is for automated testing only to validate that a representative value for any given field, matches what the typing is set in the data stream mapping. The end result, is that it will look like a union of every possible event sample in one document, yes. That is OK here.

Buildkite migration is blocked, so it's in purgatory still, will be red until a dependency is fixed. It can be safely ignored. The rest of the statuses are green.

It seems like this is entirely additions, so backporting is not out of the question for 8.11. Just ping the defend workflows team if we need to cut a release

jdu2600:
Updated the sample file (including fields added in #427).

pzl:
we should merge this to `main`, and then it's a manual backport step to `8.11` branch

jdu2600:
I've re-targeted this PR to `main`.

pzl (approved):

PR #440, opened 2023-10-05
artifacts manifest update age, snapshot date

intxgo:
## Change Summary

Add artifact snapshot date indicator. When artifacts are not pinned to any specific snapshot date the keyword 'latest' is used.

Add artifact up-to-date age indicator. It's equal to number of days since the local artifacts copy at Endpoint side was last time deemed up-to-date (the date when either Endpoint updated the artifacts or verified that it's artifacts are up-to-date with Elastic CDN).

Note, when artifacts are pinned to a specific snapshot date, the age always equal to the numbers of days since that date. In addition the initial artifacts shipped with the installation package age is equal to the number of days since the package creation date (endpoint build date).

### Sample values

Artifacts made or verified to be up-to-date today:
`"Endpoint.policy.applied.artifacts.global.update_age": 0`
Artifacts 50 days old
`"Endpoint.policy.applied.artifacts.global.update_age": 50`

Artifacts pinned to `2023-09-26` snapshot date:
`"Endpoint.policy.applied.artifacts.global.snapshot": "2023-09-26"`

Artifacts not pinned to any snapshot date:
`"Endpoint.policy.applied.artifacts.global.snapshot": "latest"`

Sample document:

Endpoint altert and Endpoint policy response:

```json
{
    "Endpoint": {
        "policy": {
            "applied": {
                "artifacts": {
                    "global": {
                        "update_age": 0,
                        "snapshot": "latest",
                    },
}

```

### For mapping changes:

- [x] I ran `make` after making the schema changes, and committed all changes
- [ ] If these field(s) are "exception"-able, I made a companion PR to Kibana adding it (see [Readme](https://github.com/elastic/endpoint-package#exceptionable))
- [ ] If this is a `metadata` change, I also updated both transform destination schemas to match

### For Transform changes:

- [ ] The new transform successfully starts in Kibana
- [ ] The corresponding transform destination schema was updated if necessary

PR #448, opened 2023-11-06
artifacts manifest update age, snapshot date (#440)

pzl:
backports #440 to `8.11`

This PR merged after the 8.11 freeze, but was never backported to the 8.11 branch.

Including it for the upcoming `8.11.1` release once #445 merges

### DIFF
--- a/custom_schemas/custom_endpoint.yml
+++ b/custom_schemas/custom_endpoint.yml
@@ -336,6 +336,16 @@
       type: object
       description: information about global protection artifacts applied.
 
+    - name: policy.applied.artifacts.global.update_age
+      level: custom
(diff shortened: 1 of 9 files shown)


kevinlog:
@pzl since we would be releasing this to a released stack version (8.11.0 is supposed to release this week), can we be sure to install this package locally first against a stack and make sure there are no issues?

pzl:
@kevinlog surely. This has been merged for a while on `main` so the 8.12 previews have had it all the while. It's also purely additive for two fields, so the expected risk is very low. But quick and easy to verify

paul-tavares (approved):

ashokaditya (approved):

PR #449, opened 2023-11-07
Bump 8.12 prerelease

pzl:
Bumps the prerelease number to force a (prerelease only) deployment of main branch


kevinlog (changes requested):
Let's hold off on the next release to 8.12.0 line until we resolve a potential issue with Serverless pulling pre-release versions

ashokaditya (approved):

PR #450, opened 2023-11-13
additional 8.11 custom documentation fields

ferullo:
## Change Summary

Add custom documentation fields that were somehow missed in my previous PR https://github.com/elastic/endpoint-package/pull/446


custom_documentation/src/endpoint/data_stream/api/windows/windows_api_credential_access.yaml
```diff
@@ -6,7 +6,7 @@ identification:
   filter:
     event.dataset: endpoint.events.api
     event.module: endpoint
-    event.type: access
+    event.provider: ObjectCallback
```
ferullo, line 9: @magermark I think this is the filtered keys we agreed on for API events and specifically the filter for credential access. Agreed?

magermark: Yup, this looks correct. Double checked the code here: https://github.com/elastic/endpoint-dev/blob/main/Plugins/CredentialProtection/Lib/CredentialAccess.cpp#L658

magermark (approved):

PR #463, opened 2024-01-05
Signing using Buildkite

gogochan:
## Change Summary

This PR adds signing using Buildkite. Publishing is not enabled yet.

Just as it was with Jenkins, signing is only enabled when branch name is either `main`, `7.*`, or `8.*`.

### Sample values
An example execution 
https://buildkite.com/elastic/endpoint-package/builds/216

### For mapping changes:

### For Transform changes:


catalog-info.yaml
```diff
@@ -21,8 +21,17 @@ spec:
     spec:
       repository: elastic/endpoint-package
       pipeline_file: ".buildkite/pipeline.yml.py"
+      branch_configuration: main 7.* 8.*
```
pzl, line 24: Is it possible to make this more restrictive? If we can avoid branches like `8.12-bugfix` or `8.13-fix-whatever` being able to release, that would be preferable.

`[78]\.[0-9]{1,2}$` or similar would be safer, if regex is possible. If not, we can figure out how to proceed and maybe lock down branches named like those examples from being created

gogochan: We can try using advanced condition.

gogochan: https://github.com/elastic/endpoint-package/pull/463/commits/996120f26b76d9a731cc6be73e933276dd1ab742

pzl (approved):

gogochan:
https://github.com/elastic/security-team/issues/6786
for tracking purpose

PR #473, opened 2024-01-17
Fix kibana version condition and additional buildkite settings

gogochan:
## Change Summary

Integration team used the same file .buildkite/scripts/find_oldest_supported_version.py, and made update to it. https://github.com/elastic/integrations/blob/main/.buildkite/scripts/find_oldest_supported_version.py

This file picks up some of the changes they introduced.

Also adding some condition logic to prevent accidental cancelation of main and 8.12 branch CI job.

### Sample values

Buildkite only changes 

### For mapping changes:

### For Transform changes:

### DIFF
--- a/.buildkite/scripts/find_oldest_supported_version.py
+++ b/.buildkite/scripts/find_oldest_supported_version.py
@@ -116,8 +116,16 @@ def run(cfg: argparse.Namespace):
     with open(cfg.manifest_path, "r") as src:
         manifest_doc = yaml.safe_load(src)
 
-    kibana_version_condition = manifest_doc["conditions"]["kibana"]["version"]
-    print(find_oldest_supported_version(kibana_version_condition), end="")
+    kibana_version_condition = ""
+    if "kibana.version" in manifest_doc["conditions"]:
+        kibana_version_condition = manifest_doc["conditions"]["kibana.version"]
+    elif "kibana" in manifest_doc["conditions"]:
+        kibana_version_condition = manifest_doc["conditions"]["kibana"]["version"]
+
+    if kibana_version_condition:
+        print(find_oldest_supported_version(kibana_version_condition), end="")
+    else:
+        print("null")
 
 
 def main():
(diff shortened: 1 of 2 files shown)


.buildkite/scripts/find_oldest_supported_version.py
```diff
@@ -116,8 +116,16 @@ def run(cfg: argparse.Namespace):
     with open(cfg.manifest_path, "r") as src:
         manifest_doc = yaml.safe_load(src)
 
-    kibana_version_condition = manifest_doc["conditions"]["kibana"]["version"]
-    print(find_oldest_supported_version(kibana_version_condition), end="")
+    kibana_version_condition = ""
+    if "kibana.version" in manifest_doc["conditions"]:
+        kibana_version_condition = manifest_doc["conditions"]["kibana.version"]
```
pzl, line 121: thanks, this will be nice if we end up having to backport further to 8.11 or something. 8.12 was where we made this yaml change

catalog-info.yaml
```diff
@@ -25,7 +25,11 @@ spec:
       # to build automatically.
       branch_configuration: main 8.12
       cancel_intermediate_builds: true
+      # Do accidently skip main or release branch that may be in the middle of releasing
+      cancel_intermediate_builds_branch_filter: '!main !8.12'
       skip_intermediate_builds: true
+      # Do accidently skip main or release branch that may be in the middle of releasing
+      skip_intermediate_builds_branch_filter: '!main !8.12'
```
pzl, line 32: can you explain what's going on here?

gogochan: I meant to say, "do not". I will update the comment. Basically `skip_intermediate_builds` cancels any currently running CI and starts a new one if the branch gets updated. For PR CI, it is not a big deal, since it doesn't involve signing and publishing. But for `main` or other releasable branch, we may not want to cancel pending publishing job because the branch was updated.

pzl: do we need to extend these things every release? e.g. when we cut an `8.13` release branch should this become `!main !8.12 !8.13`? I guess that also goes for `branch_configuration` above.

gogochan: Yes, unfortunately. The condition does not take regex afaik, and we wouldn't want to have something like `8.*`

pzl (approved):

PR #496, opened 2024-04-22
8.13 update custom documentation

intxgo:
## Change Summary

Adds document fields introduced in 8.13 and custom documentation for them.

Changes made using documents data from testing automation, where custom documentation was generated with `python3 -m endpoint.tools.update_custom_documentation`

Unfortunately there's lots of changes to "short description" fields which are unrelated to the PR. As far as I know, short description lines should not exceed 120 characters. It seems that `make clean all` reformatted the documentation to comply with the standard.

### Sample values

Sample document:

```json


```

### For mapping changes:

- [x] I ran `make` after making the schema changes, and committed all changes
- [ ] If these field(s) are "exception"-able, I made a companion PR to Kibana adding it (see [Readme](https://github.com/elastic/endpoint-package#exceptionable))
- [ ] If this is a `metadata` change, I also updated both transform destination schemas to match

### For Transform changes:

- [ ] The new transform successfully starts in Kibana
- [ ] The corresponding transform destination schema was updated if necessary

PR #497, opened 2024-04-25
8.14 update custom documentation

intxgo:
## Change Summary

Similar to https://github.com/elastic/endpoint-package/pull/496 but this time on 8.14 branch using newly developed update script https://github.com/elastic/endpoint-dev/pull/14407

~Again `make clean all` did a little change outside `custom_documentation` directory, in 2 files. It seems justified, there was a change in `user.effective.*` descriptions so it updated `fields.yaml` and `README.md`.~

I removed any changes made by `make clean all` outside `custom_documentation` directory.

### Sample values

Sample document:

```json


```

### For mapping changes:

### For Transform changes:

### DIFF
--- a/custom_documentation/doc/endpoint/alerts/windows/windows_malicious_behavior_alert.md
+++ b/custom_documentation/doc/endpoint/alerts/windows/windows_malicious_behavior_alert.md
@@ -48,8 +48,6 @@ This alert occurs when a Malicious Behavior alert occurs.
 | data_stream.dataset |
 | data_stream.namespace |
 | data_stream.type |
-| destination.ip |
-| destination.port |
 | dll.*<br /><br />dll contains dll data from the primary event in Events. It can contain any fields that any other events includes within the dll fieldset. |
 | ecs.version |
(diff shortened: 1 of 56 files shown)


intxgo:
@pzl the lack of branches with documentation was an early oversight. I'd like to keep/make them for every release.

The doc page points currently to the github repo on `main` branch. It's not ideal, but users cuould very easily switch branch from the drop-down to read the right documentation for the release they're using.

When this is merged, I'm going to cherry-pick the squashed commit into PR for main. We're planning to do the documentation checks in CI and updates on main during the development, so in the future everything should be already done before FF.

Thanks for the tip about the `clean`. I'll try that and update our procedure. The `make` generates the actual md fiiles change for the documentation. Maybe we can also replicate the behavior in our script which updates the yaml files so `make` won't be needed. I'll look into that.

intxgo:
The `clean` is needed in `make clean all` to make the md files update, but we will not commit any changes made by it outside `custom_documentation`

PR #511, opened 2024-06-18
[WIP] add file.Ext.windows.referrer_url and file.Ext.windows.host_url

AsuNa-jp:
## Change Summary

Adds new files `file.Ext.windows.referrer_url` and `file.Ext.windows.host_url` to the file event mapping. 

### Sample values

Sample document:

```json


```

## Release Target

8.15

### For mapping changes:

### DIFF
--- a/custom_schemas/custom_file.yml
+++ b/custom_schemas/custom_file.yml
@@ -73,6 +73,18 @@
       description: >
         Windows zone identifier for a file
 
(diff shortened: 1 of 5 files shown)


AsuNa-jp:
I found out that this PR is unessesary at this point. Additionally, I realized that further discussion is necessary. I might create a similar PR in the near future, but for now, I would like to close this PR since it is unnecessary now.

PR #514, opened 2024-06-27
Add `file.origin_referrer_url` and `file.origin_url` to FileEvent

AsuNa-jp:
## Change Summary

This PR adds new `file.origin_referrer_url` and `file.origin_url` to the file event mapping. These fields will also be added to the ECS. (level: extended). The following is the ECS side PR. 

* [Add new fields file.origin_referrer_url & file.origin_url](https://github.com/elastic/ecs/pull/2348)

## To reviewers

**Size of the fields**
The two fields added in this PR are intended to store URL. Therefore, the default field size of 1024 bytes is insufficient. As a result, we want to set ignore_above to 8k bytes, if possible.

* `ignore_above: 8192`

**Default Field**
This field will not necessarily be included in all file events. 
Therefore, I set it as `default_field: false`, but please let me know if this is incorrect.

**About [windows_file_open.md](https://github.com/elastic/endpoint-package/pull/514/files#diff-01a8434f3dab7b0a4176d5e984db8b91bc80f363c3a75d8f9d34ef2b24a73be8)**

As a result of running `make`, process.command_line was added to `windows_file_open.md`. If this behavior is unexpected, please let me know.

### Sample values

* file.origin_referrer_url : https://example.com
* file.origin_url : https://example.com/file.zip

## Release Target
8.15
### For mapping changes:

### DIFF
--- a/custom_documentation/doc/endpoint/file/windows/windows_file_open.md
+++ b/custom_documentation/doc/endpoint/file/windows/windows_file_open.md
@@ -63,6 +63,7 @@ This event is generated when a file is opened.
 | process.code_signature.status |
 | process.code_signature.subject_name |
 | process.code_signature.trusted |
+| process.command_line |
 | process.entity_id |
 | process.executable |
 | process.name |
--- a/custom_schemas/custom_file.yml
+++ b/custom_schemas/custom_file.yml
@@ -437,6 +437,7 @@
       type: keyword
       description: >
         The url where the file is hosted.
+
     - name: pe
       level: custom
       type: object
--- a/custom_schemas/custom_file.yml
+++ b/custom_schemas/custom_file.yml
@@ -429,12 +429,14 @@
     - name: origin_referrer_url
       level: extended
       type: keyword
+      ignore_above: 8192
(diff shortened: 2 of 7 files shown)


pzl (commented):
- no worries about the change in windows_file_open, maybe something was left behind in a previous PR? either way, things in `custom_documentation` have no ill-effect to the package functionality, so easy :+1: from me
-
>  will not necessarily be included in all file events.
> Therefore, I set it as default_field: false

I don't believe that's what that setting does, but people doing fieldless-queries against our indices is probably rare, so no harm in using that setting anyway


- can you add the sample data to `package/endpoint/data_stream/file/sample_event.json` ?  Easy :+1: after that

AsuNa-jp:
Hi @pzl 
Thank you very much for the review & feedback!
I added the sample data to `package/endpoint/data_stream/file/sample_event.json` !

pzl (approved):
:ship:

intxgo (approved):
LGTM, I'm not sure why these fields are being added there if they're going to be included in ECS

PR #523, opened 2024-07-23
Enable endpoint policy.applied.artifacts mapping

pzl:
## Change Summary

Enable `Endpoint.policy.applied.artifacts` mapping.

It has been defined for years, but set to `enabled: false`. Given that it is only a handful of fields, and very useful information to track and filter, we are enabling the mapping for search.

This is primarily for the `policy` stream, but will get pulled into `alerts` as it was already defined there

Screenshot of a filter working for a subfield:

2024-07-23-133915_scrot

And showing the field working in lens:

2024-07-23-134049_scrot

### Sample values

already included in `package/endpoint/policy/sample_event.json`

## Release Target

8.16

### For mapping changes:


joeypoon (approved):

ferullo (approved):
💯

PR #544, opened 2024-09-17
Consolidate fork,exec,exit for macOS

nicholasberlin:
## Change Summary

Changes are for aggregation of fork, exec, and exit process events for macOS.

## Release Target

8.16

### For mapping changes:

- [x] I ran `make` after making the schema changes, and committed all changes
- [ ] If these field(s) are "exception"-able, I made a companion PR to Kibana adding it (see [Readme](https://github.com/elastic/endpoint-package#exceptionable))
- [ ] If this is a `metadata` change, I also updated both transform destination schemas to match

Chained off of: https://github.com/elastic/endpoint-package/pull/543

PR #549, opened 2024-09-24
Add Windows process aggregation

nicholasberlin:
## Change Summary

Changes are for aggregation of start and exit process events for Windows.

## Release Target

8.16

### For mapping changes:

Chained off of: https://github.com/elastic/endpoint-package/pull/544


custom_documentation/doc/endpoint/process/windows/windows_process_create_and_exit.md
```diff
@@ -87,6 +87,7 @@ This event is generated when a process is created.
 | process.command_line |
 | process.entity_id |
 | process.executable |
+| process.exit_code |
```
gabriellandau, line 90: In an aggregated event, do we lose the timestamp of process termination?

pzl (commented):
if this change is only affecting the documentation, then you are good to merge.  defend workflows doesn't need to approve anything in custom docs

custom_documentation/doc/endpoint/process/windows/windows_process_create_and_exit.md
nicholasberlin: Events are aggregated in the order they are received. Aggregation entails overwriting all fields except for those marked for aggregation, which is currently only [`event.type`](https://github.com/elastic/endpoint-dev/blob/086511efc813328e9c5062977e788bc6dcfdb2bf/Python/endpoint/generators/events/v0/events.yml#L321) and [`event.action`](https://github.com/elastic/endpoint-dev/blob/086511efc813328e9c5062977e788bc6dcfdb2bf/Python/endpoint/generators/events/v0/events.yml#L360). So, `@timestamp` of the aggregated event will be the `@timestamp` from the process termination event, and the aggregated `event.action` will be `["start", "end"]`.

So, to answer your question, no. But, we do lose the `@timestamp` from the process creation. `process.start` will be overwritten as well, but I assume that `process.start` of a termination event is equal to the corresponding creation event.

gabriellandau: Okay, so it sounds like both timestamps will be present in the aggregated event.  Thanks.

gabriellandau (approved):

PR #565, opened 2024-11-26
Update schema and docs for `ptrace`, `shmget` events, fix docs for memfd events

fearful-symmetry:
## Change Summary

Still figuring out how the endpoint-package stuff works, so bare with me. 

This adds schema info for `ptrace` and `memfd` events, and also completes some of the incomplete docs that caused validation failures in the memfd branch:

```
2024-11-26 19:50:56.413 - ERROR: ecs_helper.validate_doc:553  Path process.previous.executable is not documented in /opt/buildkite-agent/builds/bk-agent-prod-gcp-1732634543267886402/elastic/endpoint-dev/ElasticCommonSchema/endpoint-package/custom_documentation/src/endpoint/data_stream/process/linux/linux_process_memfd_create.yaml
```

I'm still not entirely sure if I'm doing the custom docs correctly, it seems like the most of the `process` custom docs are copy-and-pasted, which is what I did here, but we want to keep the amount of needless doc fields to a minimum. I went through and removed a handful of the fields that were clearly specific to one event type, but a lot of them seem to originate from the event enrichment code, and I'm not sure if there's a way to verify if certain events won't be get enriched?


pzl:
FYI the CI failures are unrelated. It's due to the package already targeting 8.18 stack release, but snapshot images do not exist yet for 8.18. Ping me when this is approved and ready to merge, we bypass CI for this period of time

fearful-symmetry:
@pzl ah, alright. I assumed it was unrelated, yeah.

fearful-symmetry:
@pzl can you take a look at this? Would like to get it merged.

pzl:
@fearful-symmetry you need a reviewer from `elastic-endpoint` team @elastic/elastic-endpoint 

unfortunately that doesn't include me.. cc @ferullo

gabriellandau (commented):
Custom docs are supposed to be as minimal/streamlined as possible.  It's supposed to be a list of the fields that endpoint actually sends.  This is adding way more than the `process.previous.*` you mentioned.  Are you getting errors for all of these fields?  If not, can you trim it down to the fields for which you're receiving errors?

fearful-symmetry:
@gabriellandau I tried just adding fields as I found them, but I ran into so many sporadic document validation failures in EAF that I ended up copying most the `memfd_create` fields from the `fork_exec` fields, as my assumption was that all the  `process` events would be largely the same save for fields specific to a given `event.action` type? Seems like the different process docs are about 90% the same too. Is there an actual process for narrowing this down? I don't want to make the EAF tests even flakier than they already are.

gabriellandau:
I don't know enough about `memfd_create` events to speak to their potential fields.  Are they a subtype type of `fork_exec` event? 
 Do they set all the same fields?  I can tell you that we were given a customer requirement to document which fields Endpoint can actually send because endpoint-package's schema is way too broad, even when subsets are applied.  @ferullo created the custom documentation system to address that need.

fearful-symmetry:
@ferullo , yeah, my first pass at this was made using that too, but I ran into too many sporadic validation failures when I ran the entire EAF test suite. I did run into the session fields for the memfd events when I enabled them in the tests, for what it's worth. I'm not sure if those fields "make sense" for memfd_create events, but they're at least possible. Most of the docs in `process/linux` seem largely identical, so my assumption was coping the docs and adding the memfd-specific events was mostly a valid way to go.

fearful-symmetry:
@stanek-michal do you have any input on this? You did the first draft of it.

stanek-michal:
> @stanek-michal do you have any input on this? You did the first draft of it.

Nope, I looked into my git and notes and couldn't find anything, I think I was stuck with the exact same problem and just copied what other PRs were doing.

fearful-symmetry:
So, I got some help from other folks on how to test/update things, momentarily putting this in draft so I can test changes from endpoint...

pzl (approved):

intxgo (approved):

gabriellandau (approved):

PR #579, opened 2025-01-21
Add fleet unenrolled audit fields

pzl:
## Change Summary

Adds fleet agent unenrolled reason fields to support the new agent statuses added here: https://github.com/elastic/kibana/pull/205815

## Release Target

9.0+

### For mapping changes:

### For Transform changes:

### DIFF
--- a/package/endpoint/elasticsearch/index_template/metrics-metadata-united.json
+++ b/package/endpoint/elasticsearch/index_template/metrics-metadata-united.json
@@ -509,6 +509,12 @@
                                 "unenrollment_started_at": {
                                     "type": "date"
                                 },
+                                "audit_unenrolled_time": {
+                                    "type": "date"
+                                },
+                                "audit_unenrolled_reason": {
+                                    "type": "keyword"
+                                },
                                 "updated_at": {
                                     "type": "date"
                                 },


package/endpoint/elasticsearch/index_template/metrics-metadata-united.json
```diff
@@ -509,6 +509,12 @@
                                 "unenrollment_started_at": {
                                     "type": "date"
                                 },
+                                "audit_unenrolled_time": {
+                                    "type": "date"
+                                },
+                                "audit_unenrolled_reason": {
+                                    "type": "keyword"
+                                },
```
ashokaditya, line 517: I'm assuming these fields hold the same value as `unenrolled_at` and `unenrolled_reason`? I'm curious what triggers an update to values to these new fields.

ashokaditya (approved):

package/endpoint/elasticsearch/index_template/metrics-metadata-united.json
pzl: Not sure what went into the decision to use new fields, given those exist. But this is just mirroring the `.fleet-agents` mapping, since we union with that document. We just need to keep in sync with the agent mapping: https://github.com/elastic/elasticsearch/pull/111909/files

PR #586, opened 2025-01-31
Add 8.x release pipeline

pzl:
Release pipeline was not included for new `8.x` branch

### DIFF
--- a/.buildkite/pipeline.yml.py
+++ b/.buildkite/pipeline.yml.py
@@ -57,7 +57,7 @@ def main():
         },
     ]
 
-    if current_branch == "main" or re.match(r"^[78]\.\d+$", current_branch):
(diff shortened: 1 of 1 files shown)


.buildkite/pipeline.yml.py
```diff
@@ -57,7 +57,7 @@ def main():
         },
     ]
 
-    if current_branch == "main" or re.match(r"^[78]\.\d+$", current_branch):
+    if current_branch == "main" or current_branch == "8.x" or re.match(r"^[78]\.\d+$", current_branch):
```
ashokaditya, line 60: The regex (`^[78]\.\d+$`) should also match `8.18`.

ashokaditya (approved):

.buildkite/pipeline.yml.py
pzl: correct, the `8.18` branch will hit, but `8.x` will not.

PR #608, opened 2025-04-21
Add `zone_identifier` field to Process/DLL events

AsuNa-jp:
## Change Summary

Aligned with file.Ext.windows.zone_identifier, we've decided to add the zone_identifier field to both process and DLL events as well.

* `process.Ext.windows.zone_identifier`
* `dll.Ext.windows.zone_identifier`

### Sample values

The value of `zone_identifier` is based on the numerical values defined below.

https://learn.microsoft.com/en-us/previous-versions/windows/internet-explorer/ie-developer/platform-apis/ms537175(v=vs.85)

* Process Event
```json
            "process": {
                "Ext": {
                    "windows": {
                        "zone_identifier": 3
                    },
                },
            },
```

* DLL Event 
```json
            "dll": {
                "Ext": {
                    "windows": {
                        "zone_identifier": 3
                    },
                },
            },
```

Sample document:

sample document is added to `sample_event.json`
* [package/endpoint/data_stream/library/sample_event.json](https://github.com/elastic/endpoint-package/pull/608/files#diff-e174526294c63f77d592a363fbb4c7245654825537a7782f9dc97ab2df2f5cf8)
* [package/endpoint/data_stream/process/sample_event.json](https://github.com/elastic/endpoint-package/pull/608/files#diff-13633580ab708cec8de0486e5354f864cec2cae71df2ae2ebea41e584d8ba006)

## Release Target

9.1/8.19

### For mapping changes:

### For Transform changes:

### DIFF
--- a/.buildkite/pipeline.yml.py
+++ b/.buildkite/pipeline.yml.py
@@ -57,7 +57,7 @@ def main():
         },
     ]
 
-    if current_branch == "main" or current_branch == "8.x" or re.match(r"^[78]\.\d+$", current_branch):
+    if current_branch == "main" or current_branch == "8.x" or re.match(r"^[7-9]\.\d+$", current_branch):
         steps.append({
                 "label": "Check if published",
                 "command": ".buildkite/scripts/sign_and_publish.sh --check",
--- a/.buildkite/sign_and_publish.yml.py
+++ b/.buildkite/sign_and_publish.yml.py
@@ -20,7 +20,7 @@ def main():
     # This only gets triggered when the branch is either main or 7.\d or 8.\d
     # So, dry_run is true for non main branch
     branch = os.getenv("BUILDKITE_BRANCH")
-    dry_run = not (branch == "main" or branch == "8.x" or re.match(r"^[78]\.\d+$", branch))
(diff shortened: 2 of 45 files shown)


pzl (commented):
The changes look ok.

Can you place those sample values into the documents at `package/endpoint/data_stream/{library,process}/sample_event.json`?

Also, just to double-check the need here:

If this field needs to be searched, filtered on, or alerted on, then this PR is good. But if it doesn't have specifically those needs, and it is just a new field that endpoint is sending and you are looking to satisfy the a field-checking tool for endpoint, then you can also define this in `schema_violations` / `schema_exceptions`. You only require a mapping PR here if the field is expected to be part of a search/filter/alert query.

pzl (approved):

AsuNa-jp:
@pzl 
Thank you for approving this PR! & I've added the sample value to `package/endpoint/data_stream/{library,process}/sample_event.json`

>If this field needs to be searched, filtered on, or alerted on, then this PR is good.

Thank your sharing you thought. 
Yes, this new field need to be searched, filtered on, or alerted, so I also believe this PR is good to go.

pzl:
@AsuNa-jp  you are ok to merge :heavy_check_mark:

gabriellandau (approved):

PR #632, opened 2025-06-04
global artifacts manifest_type

intxgo:
## Change Summary

Global artifacts are rolled out gradually, only a selected few endpoints receive `candidate` artifacts to verify them before the rollout will progress. This PR adds global artifacts manifest type field to indicate which artifacts manifest type was currently used by the Endpoint.

### Sample values

Endpoint using standard stable artifacts:
`"Endpoint.policy.applied.artifacts.global.manifest_type": "stable"`

Endpoint testing candidate artifacts:
`"Endpoint.policy.applied.artifacts.global.manifest_type": "candidate"`

Sample document:

Endpoint altert and Endpoint policy response:

```json
{
    "Endpoint": {
        "policy": {
            "applied": {
                "artifacts": {
                    "global": {
                        "manifest_type": "stable",
                    },
}

```

### For mapping changes:

### For Transform changes:


custom_documentation/src/endpoint/data_stream/metrics/metrics.yaml
```diff
@@ -103,8 +103,8 @@ fields:
   - Endpoint.metrics.memory.endpoint.private.latest
   - Endpoint.metrics.memory.endpoint.private.mean
   - Endpoint.metrics.queue_metrics.kernel.send_to_user_queue.drops
-  - Endpoint.metrics.queue_metrics.kernel.send_to_user_queue.size
   - Endpoint.metrics.queue_metrics.kernel.send_to_user_queue.last_drop_time_utc
+  - Endpoint.metrics.queue_metrics.kernel.send_to_user_queue.size
```
intxgo, line 107: it looks like some of the the custom documentation was modified manually, so the update script moved things around when sorting the updated filed

⚠️ custom documentation should be updated by script ideally (on Linux, as it's the only "approved" system for running `make` in endpoint-package repository). Updating it manually is not an error, but it leads to unexpected changes on other's PRs, such as unsorted fields

PR #637, opened 2025-06-12
Add new policy fields for firewall_anti_tamper plugin

matthewh-elastic:
## Change Summary

Add additional Policy field entries for new `firewall_anti_tamper` plugin.

### For mapping changes:

### For Transform changes:

### DIFF
--- a/custom_schemas/custom_endpoint.yml
+++ b/custom_schemas/custom_endpoint.yml
@@ -171,24 +171,6 @@
         but not a simple sum of the actions
       short: the overall status of memory_protection
 
-    - name: policy.applied.response.configurations.firewall_anti_tamper
-      level: custom
-      type: object
-      description: overall firewall anti-tamper configuration and status of the applied policy
-
-    - name: policy.applied.response.configurations.firewall_anti_tamper.concerned_actions
-      level: custom
-      type: keyword
-      description: all actions that were taken for firewall anti-tamper
-
-    - name: policy.applied.response.configurations.firewall_anti_tamper.status
-      level: custom
-      type: keyword
-      description: >
-        the overall status of firewall anti-tamper, this is correlated to the status of concerned actions
-        but not a simple sum of the actions
-      short: the overall status of firewall anti-tamper
-
     - name: policy.applied.response.configurations.streaming
       level: custom
       type: object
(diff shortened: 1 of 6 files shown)


tomsonpl (approved):

package/endpoint/data_stream/policy/fields/fields.yml
```diff
@@ -245,6 +245,23 @@
       ignore_above: 1024
       description: the overall status of event collection, this is correlated to the status of concerned actions  but not a simple sum of the actions
       default_field: false
+    - name: policy.applied.response.configurations.firewall_anti_tamper
+      level: custom
+      type: object
+      description: overall firewall_anti_tamper configuration and status of the applied policy
```
intxgo, line 251:
```suggestion
      description: overall firewall anti-tamper configuration and status of the applied policy
```

Looking at other descriptions, we never use variable (node) names to explain the variable

intxgo (commented):
the descriptions doesn't look good

package/endpoint/data_stream/policy/fields/fields.yml
matthewh-elastic: Good spot, it looks like `behavior_protection` and `memory_protection` use the same format at this, but most use plain english. I'll update this to "firewall anti-tamper".

matthewh-elastic: Updated, thanks 👍

intxgo (approved):

custom_schemas/custom_endpoint.yml
```diff
@@ -171,6 +171,24 @@
...
+      description: all actions that were taken for firewall anti-tamper
+
+    - name: policy.applied.response.configurations.firewall_anti_tamper.status
+      level: custom
+      type: keyword
+      description: >
+        the overall status of firewall anti-tamper, this is correlated to the status of concerned actions
+        but not a simple sum of the actions
+      short: the overall status of firewall anti-tamper
+
```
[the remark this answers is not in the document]
matthewh-elastic: Thanks. All of the fields for existing plugins are currently in the same format (eg [here](https://github.com/elastic/endpoint-package/blob/main/custom_schemas/custom_endpoint.yml#L88)) so I copied these for consistency. Perhaps a separate PR could be raised to add the `example` field to all of them at once?

PR #642, opened 2025-06-18
Add tags to action request documents

pzl:
## Change Summary

Adds field `tags` to actions

## Release Target

9.1

### For mapping changes:

### DIFF
--- a/custom_subsets/elastic_endpoint/actions/actions.yaml
+++ b/custom_subsets/elastic_endpoint/actions/actions.yaml
@@ -31,6 +31,7 @@ fields:
   space:
     fields:
       originSpaceId: {}
+      tags: {}
   rule:
     fields:
       id: {}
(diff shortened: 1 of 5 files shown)


custom_subsets/elastic_endpoint/actions/actions.yaml
```diff
@@ -31,6 +31,7 @@ fields:
   space:
     fields:
       originSpaceId: {}
+      tags: {}
```
paul-tavares, line 34: Whats this `space` field?  Are these being defined as `space.originSpaceId` and `space.tags` on the index?  If so, that does not line up with what we discussed.

pzl: not a field, just a "category" so it knows where to pull the tags definition from.

`tags` is a top-level field. you can take a look at the change in `sample_event.json` to see the end result

paul-tavares: Got it. Thanks.

PR #660, opened 2025-08-18
Add device schema

matthewscherer:
## Change Summary
This adds a new schema for device documents. Currently we only support volume devices but more will follow.

### Sample values

Sample document:
A sample document has been checked into the repo with this PR.
`package/endpoint/data_stream/device/sample_event.json`

## Release Target

9.2.0


matthewscherer:
> Looks good except `writable` is mispelled `writeable`.
> 
> I noticed you're mapping more fields than Endpoint produces. Is that on purpose?

Nice catch. Thank you.

PR #663, opened 2025-09-02
update the custom documentation README

ferullo:
This updates the README for custom documentation with information about which Kibana features are associated with each datastream.


gergoabraham (commented):
i don't really know if this is semantically correct, but checked it syntactically : )

nicholasberlin (approved):

PR #668, opened 2025-09-08
Added the new device fields.

matthewscherer:
## Change Summary

This will add new fields to the device key that supports the new device control plugin.

### Sample values
A sample was provided in the example json file. 

## Release Target
9.2.0


ferullo (approved):
LGTM this matches https://github.com/elastic/ecs/pull/2530

PR #670, opened 2025-09-09
Added pe architecture to alerts

bit-envoy:
## Change Summary

Added PE architecture for malware alerts.

### Sample values
x64, x86, arm64, ia64, unknown

Sample document:

```json
        "pe": {
            "file_version": "0.0.0.0",
            "description": " ",
            "original_file_name": "5t5mpwxc.dll",
            "architecture": "x64"
        }
```

### For mapping changes:

### For Transform changes:


bit-envoy:
> @bit-envoy fyi the build failures are pretty much just about formatting. the linter tool is unreasonably picky about the `sample_event.json` file _not_ ending in a newline.
> 
> and the custom doc markdown file -- I'm guessing that's an alphabetical-order field order thing. The build step essentially just runs `make` (which includes a **format** step) and then checks if there is any diff from your branch. So the format step is catching those two things I figure

How am I suppose to sort this file? It does not look sorted for me. (e.g. file.pe.Ext.dotnet should be after file.pe.company)

pzl (approved):

PR #720, opened 2026-01-28
Add hit/miss metrics fields for trusted ancestors reporting

fearful-symmetry:
## Change Summary

This adds two custom fields for reporting trusted ancestor lookup hits/misses:

```
  - Endpoint.metrics.events_cache.trusted_ancestors.hits
  - Endpoint.metrics.events_cache.trusted_ancestors.misses
```

I've never had to add anything to endpoint-package, so I'm not 100% sure I have everything completed here, even though I've run `make clean all`.

If you have alternate opinions about how these fields should be named, please tell me, I just came up with these in about 4 seconds.

### Sample values

```
    "events_cache": {
        "trusted_ancestors": {
            "hits": 504,
            "misses": 23578
        }
    },
```

## Release Target

9.x

### For mapping changes:

### For Transform changes:

### DIFF
--- a/custom_documentation/doc/endpoint/metrics/metrics.md
+++ b/custom_documentation/doc/endpoint/metrics/metrics.md
@@ -101,6 +101,8 @@ This is an internal state management document that includes metrics on Endpoint'
 | Endpoint.metrics.documents_volume.volume_device_events.suppressed_count |
 | Endpoint.metrics.event_filter.active_global_count |
 | Endpoint.metrics.event_filter.active_user_count |
+| Endpoint.metrics.events_cache.trusted_ancestors.hits |
+| Endpoint.metrics.events_cache.trusted_ancestors.misses |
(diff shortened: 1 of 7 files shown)


gabriellandau (changes requested):
Can you add these fields to the [sample metric document](https://github.com/elastic/endpoint-package/blob/main/package/endpoint/data_stream/metrics/sample_event.json) for the unit tests?

gabriellandau (approved):

intxgo (approved):

nicholasberlin (approved):

pzl (approved):
Change is good, thank you for including values in `sample_event.json`.

If `double` is the appropriate data type here, then you are ok to merge

fearful-symmetry:
@pzl , yeah, I agree, unsigned_long is probably more appropriate here, I agree.

PR #721, opened 2026-02-02
Change trusted_ancestor types from double to unsigned_long

fearful-symmetry:
## Change Summary

CC @pzl 

This is a minor update to  https://github.com/elastic/endpoint-package/pull/720 that changes the metrics types from double to unsigned_long, which is more appropriate. 

### For mapping changes:


pzl (approved):
Looks great, thanks for following up on this!

PR #728, opened 2026-03-13
automatic troubleshooting endpoint context docs

joeypoon:
## Change Summary

Adds Automatic Troubleshooting context docs useful for troubleshooting common endpoint issues.


pzl (approved):

ferullo (commented):
There's a lot to review here so I'm going to submit a review for each file. The first one is done. It's reasonable/expected to sit on my recommendations until an Endpoint developer reviews too (so they can see my comments and contradict me, etc).

package/endpoint/docs/knowledge_base/high_cpu/linux_high_cpu.md
```diff
@@ -0,0 +1,102 @@
...
+For Kafka outputs, `Message size too large` errors cause repeated delivery failures. On 8.18.3+, oversized messages are dropped gracefully instead of retried indefinitely.
+
+### Events plugin hung hashing large binaries during policy application
+
+During policy application, Elastic Defend hashes all running processes and their executables. When the file cache (`/opt/Elastic/Endpoint/state/cache.db`) is empty — first install, cache deleted, or after upgrade — the endpoint must hash every binary from scratch. Large binaries (e.g. Oracle at `/data/app/oracle/product/*/bin/oracle`) can cause the Events plugin to hang in `ConfigurationCallback` while performing SHA1 hashing, driving CPU to 100% for extended periods.
+
+The endpoint will report status as CONFIGURING during this time:
+- `Endpoint is setting status to CONFIGURING, reason: Policy Application Status`
+
+On first run with an empty cache, the CONFIGURING phase can take 5–30 minutes depending on the number and size of running processes. This is expected behavior. Subsequent restarts are fast because the cache persists.
```
nicholasberlin: 5-30 minutes? is that supposed to be seconds?

package/endpoint/docs/knowledge_base/endpoint_exceptions/endpoint_exceptions_not_working.md
```diff
@@ -0,0 +1,92 @@
+---
+type: automatic_troubleshooting
+sub_type: endpoint_exceptions
+link: https://www.elastic.co/docs/solutions/security/manage-elastic-defend/optimize-elastic-defend
+date: '2026-03-11'
+---
+
+## Symptom
+
+An Endpoint Alert Exception has been created in Elastic Defend, but the endpoint continues to generate alerts or block processes that should be excluded. Alert documents matching the exception criteria still appear in `logs-endpoint.alerts-*`.
```
ferullo, line 10: We should add another doc for when Alert documents do not appear in `logs-endpoint.alerts-*`
* A shipping error from Endpoint
* A configuration error in Kafka or Logstash
* Actually Endpoint isn't actively blocking activity, it's caused by some sort of interaction with another AV or process -- targeted Trusted Apps is the most common solution

package/endpoint/docs/knowledge_base/endpoint_exceptions/endpoint_exceptions_not_working.md
```diff
@@ -0,0 +1,92 @@
...
+- **Wrong field name**: Using `file.path` (the file being acted upon) when the alert populates `process.executable` (the running binary), or vice versa. Check the actual alert document in `logs-endpoint.alerts-*` to confirm which fields are populated and what values they contain.
+- **Keyword vs text field type**: If the exception references a field mapped as `text` (analyzed, tokenized) rather than `keyword` (exact value), the match behavior may differ. Exception entries work against the raw field values stored by the endpoint, so verify the field mapping in the alert index.
+
+To diagnose: query `logs-endpoint.alerts-*` for a recent alert that should have been suppressed. Compare each field and value in the exception entry against the actual alert document fields character-by-character.
+
+### Exception scoped to wrong integration policy
+
+Endpoint Alert Exceptions can be scoped to specific integration policies ("per policy") or applied globally. If the exception is assigned to a policy that does not cover the affected endpoint, it has no effect on that endpoint.
```
ferullo: This is not true. We plan to ship per-policy exceptions in 9.4 and that'll be opt-in.

package/endpoint/docs/knowledge_base/endpoint_exceptions/endpoint_exceptions_not_working.md
```diff
@@ -0,0 +1,92 @@
...
+
+1. Create a role with `all` privileges on `.kibana_security_solution*` with `allow_restricted_indices: true`.
+2. As a user with that role, run: `PUT /.kibana_security_solution/_settings { "index": { "mapping": { "nested_objects": { "limit": 20000 } } } }`.
+3. Monitor Kibana logs for "Committed manifest" messages, which indicate the manifest task is recovering and distributing artifacts again.
+
+If the limit is still exceeded after raising it to 20000, increase further in increments of 10000. Deployments with hundreds of policies and thousands of artifact entries may need 30000 or higher.
+
+### Exception on wrong cluster in cross-cluster search (CCS) setups
+
+In architectures where Fleet-managed agents send data to a "data cluster" and analysts query via CCS from a separate "analyst cluster", Endpoint Alert Exceptions must be created on the data cluster (the one running Fleet Server). Exceptions created on the analyst cluster are not distributed to endpoints because CCS does not support Fleet Actions.
```
ferullo:
```suggestion
In some architectures Fleet-managed agents are managed via an "analyst cluster" but write bulk data to a "data cluster".. In these architectures, analysts query data via CCS from the data cluster, Endpoint Alert Exceptions must be created on the analyst cluster (the one running Fleet Server). Exceptions created on the data cluster are not distributed to endpoints because CCS does not support Fleet Actions.
```

package/endpoint/docs/knowledge_base/endpoint_exceptions/endpoint_exceptions_not_working.md
```diff
@@ -0,0 +1,92 @@
...
+
+If exceptions exist only on the analyst cluster, alert documents will continue to be generated by endpoints and indexed on the data cluster. The analyst cluster will then alert on those documents via CCS.
```
ferullo:
```suggestion
If exceptions exist only on the data cluster, alert documents will continue to be generated by endpoints and indexed on the data cluster. The analyst cluster will then alert on those documents via CCS.
```

package/endpoint/docs/knowledge_base/endpoint_exceptions/endpoint_exceptions_not_working.md
```diff
@@ -0,0 +1,92 @@
...
+
+### Exception not suppressing Kibana detection rule alerts
+
+Endpoint Alert Exceptions suppress alerts generated by the endpoint itself — they prevent the endpoint from creating alert documents in `logs-endpoint.alerts-*`. However, if a Kibana detection rule (e.g. the prebuilt "Malware Detection Alert" rule) is configured to fire on endpoint alert documents, that Kibana rule generates its own separate alert documents.
+
+If the endpoint exception is working correctly (no new documents in `logs-endpoint.alerts-*`), but alerts still appear in the Kibana Alerts page, the source is the Kibana detection rule, not the endpoint. In this case, add a Detection Rule Exception on the Kibana rule itself, or disable the redundant detection rule.
+
+To distinguish: check the `kibana.alert.rule.rule_type_id` field. Endpoint-generated alerts have `event.module: endpoint` and `event.dataset: endpoint.alerts`. Kibana detection rule alerts have fields like `kibana.alert.rule.category: "Custom Query Rule"` and `kibana.alert.rule.producer: "siem"`.
```
ferullo: This whole section makes no sense. It's not possible for the SIEM rule to create an alert if there is no underlying Endpoint alert in `logs-endpoint.alerts-*`.

Maybe we mean to say alerts created by most SIEM rules need SIEM exceptions not Endpoint exceptions?

package/endpoint/docs/knowledge_base/endpoint_exceptions/endpoint_exceptions_not_working.md
```diff
@@ -0,0 +1,92 @@
...
+
+### Protection mode confusion: detect vs prevent
+
+When malware protection is set to "detect" mode, the endpoint generates alert documents but does not block execution. If a user reports that a legitimate application is being "blocked" while in detect mode, the blocking may come from a different protection feature:
+
+- **Attack Surface Reduction (credential hardening)**: This protection does not have a "detect" mode — it always prevents. If `attack_surface_reduction.credential_hardening.enabled` is `true` in the policy, it can block installers and other processes regardless of the malware protection mode setting.
+- **Malware On Write scanning**: Files can be quarantined or blocked during write operations even in detect mode if another protection layer (e.g. memory protection in prevent mode) intervenes.
```
ferullo:
```suggestion
```

ferullo:
```suggestion
- **Inter-Application Incompatibility**: Sometimes it seems like Endpoint is blocking system activity because of a false positive but actually its caused by an incompatibility between Endpoint and another application on the host. If alerts are being ingested into Elasticsearch in other scenarios from the same host this points to an incompatibility as the underlying cause. A targeted Trusted App entry is a common solution. If Malware protection is enabled, use `curl https://secure.eicar.org/eicar.com.txt --output [eicar.com](http://eicar.com/).txt` to test if alerts are being ingested into Elasticsearch.
```

Maybe my note on testing with EICAR should go somewhere else?

package/endpoint/docs/knowledge_base/endpoint_exceptions/false_positive_malware_detection.md
```diff
@@ -0,0 +1,95 @@
...
+
+### Behavioral rule false positive
+
+Behavioral detection rules monitor runtime patterns across the system. A false positive occurs when legitimate software performs actions that match a malicious pattern (e.g. ransomware-like file encryption, unusual DLL loading via `rundll32.exe`).
+
+The alert's `kibana.alert.rule.name` or `rule.description` field identifies which behavioral rule fired. Common false-positive behavioral rules include:
+- "Unusual DLL Extension Loaded by Rundll32 or Regsvr32" — can fire during legitimate software installations that use `rundll32.exe` to load DLLs from temporary directories.
+- "Suspicious PowerShell Execution" — can fire for legitimate IT management scripts.
+
+To suppress on Elastic Defend 9.2+: add the process as a **Trusted Application** (this disables behavioral detections for the process entirely) OR add an **Endpoint Alert Exception** targeting the specific rule ID and process if you want to suppress only that one rule.
```
ferullo:
```suggestion
To suppress: add an **Endpoint Alert Exception** targeting the specific rule ID and process if you want to suppress only that one rule.
```

package/endpoint/docs/knowledge_base/endpoint_exceptions/false_positive_malware_detection.md
```diff
@@ -0,0 +1,95 @@
...
+
+To suppress on pre-9.2: Trusted Applications do not suppress behavioral detections on these older versions. Use an **Endpoint Alert Exception** targeting `rule.id` and the relevant process fields.
```
ferullo:
```suggestion
```

package/endpoint/docs/knowledge_base/endpoint_exceptions/false_positive_malware_detection.md
```diff
@@ -0,0 +1,95 @@
...
+
+### Memory signature false positive
+
+Memory signature alerts (`event.code: memory_signature`) occur when YARA rules match patterns found in executable memory regions. Rare false positives can occur when:
+- The endpoint's own process memory contains YARA signature strings (because the signatures themselves are loaded into memory for matching).
+- A third-party driver or application causes memory corruption that makes non-executable memory pages appear executable, leading to unexpected scanning of heap regions.
+- Faulty RAM causes bit flips in page table entries, changing page protection bits and exposing non-code memory to the scanner.
```
ferullo:
```suggestion
```

package/endpoint/docs/knowledge_base/endpoint_exceptions/false_positive_malware_detection.md
```diff
@@ -0,0 +1,95 @@
...
+
+In rare cases, the endpoint may populate alert fields with corrupted data (e.g. truncated `dll.path` or `dll.name` values due to kernel-level memory corruption from a third-party driver). When the corrupted field values happen to match a behavioral detection rule's criteria — for example, a DLL path truncated to appear as if it has an unusual extension — the rule fires on data that does not reflect the actual system state.
+
+These are product bugs, not fixable via exceptions in the general case. However, a targeted Endpoint Alert Exception can suppress the specific false-positive pattern as a workaround. For example, suppressing a rule when the DLL's code signature status indicates it was never actually loaded (`dll.code_signature.status: "errorCode_endpoint: Initital state, no attempt to load signature was made"`).
+
+If you encounter alerts where field values appear truncated or nonsensical, collect agent diagnostics and the full alert document for Elastic Support. Check for third-party kernel drivers that may be causing memory corruption — the Windows Driver Verifier can help identify misbehaving drivers on non-production systems.
```
ferullo: This was a very edge case, I think we should remove it. It won't generalize well.
```suggestion
```

package/endpoint/docs/knowledge_base/endpoint_exceptions/false_positive_malware_detection.md
```diff
@@ -0,0 +1,95 @@
...
+
+### Blocklist entry not blocking a trusted-signed process
+
+The Elastic Defend blocklist prevents known-malicious binaries from executing. However, if a process matches the **Global Exception List** before the blocklist is evaluated, the blocklist entry is skipped. The Global Exception List contains entries for major trusted software publishers (e.g. Microsoft). This means binaries signed by these vendors cannot be blocked via the user blocklist — the global exception takes priority.
+
+This is a known product limitation. If the goal is to prevent specific trusted-signed binaries from running, the blocklist is not the correct mechanism. Consider using operating system-level application control policies (e.g. Windows AppLocker or WDAC) instead.
```
ferullo:
```suggestion
```

package/endpoint/docs/knowledge_base/endpoint_exceptions/false_positive_malware_detection.md
```diff
@@ -0,0 +1,95 @@
...
+
+
+## Investigation priorities
+
+1) Determine the detection engine that generated the alert by checking `event.code`: `malicious_file` (malware engine), `behavior` (behavioral protection), `memory_signature` (memory scanning). This determines which artifact type to use for suppression.
```
ferullo:
```suggestion
1) Determine the detection engine that generated the alert by checking `event.code`: `malicious_file` (malware protection), `behavior` (malicious behavior protection), `memory_signature` (memory threat protection), `ransomware` (ransomware protection). This determines which artifact type to use for suppression.
```

package/endpoint/docs/knowledge_base/high_cpu/linux_high_cpu.md
```diff
@@ -0,0 +1,102 @@
...
+Use `sudo elastic-endpoint top` on the affected host to identify which processes and internal processing stages consume the most CPU. Query `metrics-endpoint.metrics-*` for `Endpoint.metrics.system_impact` to identify the top processes remotely.
+
+
+## Common issues
+
+### Monitoring scripts spawning many short-lived child processes
+
+Monitoring and automation scripts that run on a schedule (cron, systemd timers) and spawn many child processes are the most common cause of high CPU on Linux. A single monitoring script invoking `curl`, `mysql`, `ssh`, `grep`, `sed`, `awk`, and `bash` in rapid succession generates a burst of process creation events, each of which Elastic Defend must enrich and evaluate against behavioral rules.
+
+A typical pattern is hourly CPU spikes lasting 5–10 minutes, aligning with cron schedules (e.g. xx:31–xx:41 every hour). In one case, a script at `/var/cache/system-monitoring/helper/compare-inventory.sh` that used `curl` to collect data triggered the behavioral rule "Suspicious Download and Redirect by Web Server" 86,340 times in a single diagnostics window, driving endpoint service CPU above 200% (out of 800% on 8 cores).
```
ferullo:
```suggestion
A typical pattern is hourly CPU spikes lasting 5–10 minutes, aligning with cron schedules (e.g. xx:31–xx:41 every hour).
```

package/endpoint/docs/knowledge_base/high_cpu/linux_high_cpu.md
```diff
@@ -0,0 +1,102 @@
...
+
+Adding the parent script as a Trusted Application stops monitoring of its process tree but does not prevent behavioral rules from firing if the rule matches on child process characteristics. On versions prior to 9.2, behavioral detections still fire for trusted processes. On 9.2+, behavioral detections are disabled for Trusted Applications.
```
ferullo: @nicholasberlin Trusted Application Descendants or Event Filter Descendants will help here, right?

package/endpoint/docs/knowledge_base/high_cpu/linux_high_cpu.md
```diff
@@ -0,0 +1,102 @@
...
+
+Remediations:
+- Create an **Endpoint Alert Exception** targeting the specific rule ID and parent process:
+  - `rule.id IS <rule-id>` AND `process.parent.executable IS /path/to/script`
+- Upgrade to 8.19.11+ or 9.2+ for improved handling of trusted process behavioral rules.
+- Review `linux.advanced.events` settings to disable unnecessary event types (e.g. `linux.advanced.events.dns` if DNS monitoring is not needed).
```
ferullo:
```suggestion
- Review `linux.advanced.events` settings to disable unnecessary event types (e.g. `linux.advanced.events.dns` if malicious behavior rules based on DNS monitoring is not needed).
```

package/endpoint/docs/knowledge_base/high_cpu/linux_high_cpu.md
```diff
@@ -0,0 +1,102 @@
...
+- Use **Event Filters** to reduce event volume from known-noisy directories without creating a monitoring blind spot.
```
ferullo:
```suggestion
- Use **Event Filters** to reduce event volume from known-noisy directories without creating a active protection blind spot.
```

package/endpoint/docs/knowledge_base/high_cpu/linux_high_cpu.md
```diff
@@ -0,0 +1,102 @@
...
+
+CPU returns to normal within approximately 40 seconds after connectivity is restored. The command `sudo /opt/Elastic/Endpoint/elastic-endpoint test output` can be used to verify output connectivity — on affected versions this command itself will spike CPU when the output is unreachable.
+
+Upgrade to 8.13.4+ where the retry loop includes proper backoff. Check `logs-elastic_agent.endpoint_security-*` for the error patterns above.
+
+For Kafka outputs, `Message size too large` errors cause repeated delivery failures. On 8.18.3+, oversized messages are dropped gracefully instead of retried indefinitely.
+
+### Events plugin hung hashing large binaries during policy application
+
+During policy application, Elastic Defend hashes all running processes and their executables. When the file cache (`/opt/Elastic/Endpoint/state/cache.db`) is empty — first install, cache deleted, or after upgrade — the endpoint must hash every binary from scratch. Large binaries (e.g. Oracle at `/data/app/oracle/product/*/bin/oracle`) can cause the Events plugin to hang in `ConfigurationCallback` while performing SHA1 hashing, driving CPU to 100% for extended periods.
```
ferullo:
```suggestion
During policy application, Elastic Defend hashes all running processes and their executables. When the file cache (`/opt/Elastic/Endpoint/state/cache.db`) is empty — first install, cache deleted, or after upgrade — the endpoint must hash every binary from scratch. Large binaries (e.g. Oracle at `/data/app/oracle/product/*/bin/oracle`) can drive CPU to 100% for extended periods.
```

package/endpoint/docs/knowledge_base/trusted_apps/trusted_apps_not_effective.md
```diff
@@ -0,0 +1,77 @@
+---
+type: automatic_troubleshooting
+sub_type: trusted_apps
+link: https://www.elastic.co/docs/solutions/security/manage-elastic-defend/optimize-elastic-defend
```
tomsonpl, line 4: I see that only 3 files contain the `link` field, just checking if this is intentional?

package/endpoint/docs/knowledge_base/incompatible_software/incompatible_software_third_party.md
```diff
@@ -0,0 +1,90 @@
+---
+type: automatic_troubleshooting
+sub_type: incompatible_software
+date: '2026-03-11'
```
tomsonpl, line 5: no OS field here?

tomsonpl (approved):
Looks alright, left a few questions just for clarity sake :)

package/endpoint/docs/knowledge_base/high_cpu/windows_high_cpu.md
```diff
@@ -0,0 +1,96 @@
...
+---
+
+## Symptom
+
+Elastic Defend's `elastic-endpoint.exe` process consumes sustained high CPU on Windows hosts, causing system slowdown, application unresponsiveness, or login delays. CPU may spike during specific activities (application launches, user logons, Office updates) or build gradually over hours until the system becomes unusable.
+
+
+## Summary
+
+Elastic Defend performs real-time file hashing, digital signature verification, memory scanning, behavioral rule evaluation, and event enrichment. Each protection layer adds processing overhead, and the cumulative effect depends on the workload profile of the host. The most common drivers of high CPU on Windows are third-party security product conflicts creating mutual scanning loops, high-volume security events (logon/logoff) overwhelming the rules engine, file-intensive operations triggering repeated hashing of large binaries, and VDI/Citrix environments where the file metadata cache is empty on every session.
```
ferullo:
```suggestion
Elastic Defend performs real-time file hashing, digital signature verification, memory scanning, behavioral rule evaluation, and event enrichment. Each protection layer adds processing overhead, and the cumulative effect depends on the workload profile of the host. The most common drivers of high CPU on Windows are third-party security product conflicts creating mutual scanning loops, high-volume security events (logon/logoff) overwhelming the Malicious Behavior engine, file-intensive operations triggering repeated hashing of large binaries, and VDI/Citrix environments where the file metadata cache is empty on every session.
```

package/endpoint/docs/knowledge_base/high_cpu/windows_high_cpu.md
```diff
@@ -0,0 +1,96 @@
...
+- On older versions, add `C:\Windows\System32\lsass.exe` as a Trusted Application to drop logoff events (attributed to lsass). This reduces visibility into lsass file and registry activity.
+- Disabling Malicious Behavior Protection stops Security event collection entirely but removes behavioral detection capability.
+- Removing "Security" from the Event Collection settings in the policy does not help while Malicious Behavior Protection is enabled, since behavioral rules require these events.
+
+### File-intensive operations and large binary hashing
+
+Elastic Defend hashes and verifies digital signatures of executables and DLLs when they are loaded. Large binaries like `msedge.dll` (195 MB) can take 10–15 seconds per hash operation. On hosts running browsers, Office applications, or developer tools that load many large DLLs, this creates sustained CPU spikes.
+
+The endpoint maintains a file metadata cache to avoid re-hashing known files. On versions prior to 8.0.1, the cache size (`FILE_OBJECT_CACHE_SIZE`) was limited to 500 entries, leading to frequent cache evictions and redundant hashing. Upgrading to 8.0.1+ significantly improves caching behavior.
```
ferullo: pre 8.0.1 is so old it is no longer supported.

```suggestion
```

package/endpoint/docs/knowledge_base/high_cpu/windows_high_cpu.md
```diff
@@ -0,0 +1,96 @@
...
+
+For immediate relief on older versions, set `windows.advanced.kernel.asyncimageload` and `windows.advanced.kernel.syncimageload` to `false` in advanced policy settings. This reduces CPU by approximately 85% for DLL load processing at the cost of reduced visibility into library load events.
```
ferullo:
```suggestion
```

package/endpoint/docs/knowledge_base/bsod/windows_bsod_endpoint_driver.md
```diff
@@ -0,0 +1,80 @@
...
+
+**Fixed version**: 9.2.4.
+
+**Mitigation**: Downgrade to the prior agent version (e.g. 9.2.1) until the upgrade to 9.2.4+ can be performed.
+
+### Third-party kernel driver conflicts
+
+Other security products running kernel-mode drivers can interfere with Elastic Defend's driver initialization or runtime operation. The most commonly reported conflicts include:
+
+- **Trellix Access Control**: Trellix's kernel driver can intercept the Windows Base Filtering Engine (BFE) service, causing Defend's WFP (Windows Filtering Platform) driver initialization to hang or take an extremely long time. When another thread within Defend tasks the kernel driver before initialization completes, the system crashes. The call stack typically shows `elastic_endpoint_driver!HandleIrpDeviceControl` calling `bkRegisterCallbacks` with `KeAcquireGuardedMutex` on an uninitialized mutex. This interaction was introduced by a WFP driver refactor in 8.16.0. Fixed in 8.17.6, 8.18.1, and 9.0.1. Workaround: disable Trellix Access Protection or add a Trellix exclusion for the BFE service (`svchost.exe`).
```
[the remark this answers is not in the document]
bjmcnic: I don't think that's a good recommendation to give a customer after the fixed versions were released.  At this point the recommendation should be to upgrade.

package/endpoint/docs/knowledge_base/trusted_apps/trusted_apps_not_effective.md
```diff
@@ -0,0 +1,77 @@
...
+---
+
+## Symptom
+
+A Trusted Application has been configured in Elastic Defend, but the endpoint still appears to monitor the trusted process, consume high CPU for it, or generate alerts related to it. The process may still appear in `elastic-endpoint top` output, or malware/behavioral alerts continue to fire.
+
+
+## Summary
+
+A Trusted Application tells Elastic Defend to stop monitoring a process entirely — it creates an intentional blind spot. This is fundamentally different from an Endpoint Alert Exception, which only suppresses alerts while continuing to monitor. Many "trusted app not working" issues stem from using the wrong artifact type, misconfiguring the entry's condition fields, or running a version affected by known event-processing bugs.
```
intxgo, line 15:
```suggestion
A Trusted Application tells Elastic Defend to stop monitoring a process entirely — it creates an intentional blind spot - implicitly meaning it only works on executables. This is fundamentally different from an Endpoint Alert Exception, which only suppresses alerts while continuing to monitor. Many "trusted app not working" issues stem from using the wrong artifact type, misconfiguring the entry's condition fields, or running a version affected by known event-processing bugs.
```

package/endpoint/docs/knowledge_base/trusted_apps/trusted_apps_not_effective.md
```diff
@@ -0,0 +1,77 @@
...
+Key distinction:
+- **Trusted Application**: Stops Endpoint from *monitoring* the process. Use only for software conflicts or performance issues where full bypass is acceptable.
+- **Endpoint Alert Exception**: Stops Endpoint from *generating alerts or blocking processes*. Use for false positives where you still want monitoring.
+
+Even when working correctly, Trusted Applications still generate process events for visualizations and internal Elastic Stack use. Some CPU usage from a trusted process is expected.
+
+
+## Common issues
+
+### Wrong condition field or value
```
intxgo, line 34:
```suggestion
## Common issues

### Logical misuse

Trusted application makes an application process trusted, allowing it to do whatever it does, modify files, start other applications, etc. When you're creating a Trusted Application entry you're adding a filter telling {{elastic-defend}} to fully trust a process matching the filter and all its actions. However it can't be used the other way around, to whitelist a particular action for some process(es), for example to trust any app to modify a particular csv file whilst doing full monitoring for other actions mabe by  the process(es).

### Wrong condition field or value
```

or `### Conceptual misues` ?

I can see from SDHs that people really misuse TA, most likely because we don't have a commonly used feature "Suppress protections (trust) everything matching the path: xxx".

The problem is that they hardly ever cleanup unsuccessful TA entries, either they don't know or don't care, but the more entries the more CPU is spend on every process start evaluating the TA list.

We've seen TA artifacts containing paths like `C:\somethin\*.sys`, `C:\*\*.txt`

package/endpoint/docs/knowledge_base/missed_checkins/windows_missed_checkins.md
```diff
@@ -0,0 +1,88 @@
+---
+type: automatic_troubleshooting
+sub_type: missed_checkins
+os: [Windows]
+date: '2026-03-09'
+---
+
+## Symptom
+
+Elastic Agent reports the Endpoint component as `FAILED` or `DEGRADED` with the message "Failed: endpoint service missed 3 check-ins". The Endpoint may appear unhealthy in Fleet, and endpoint events, alerts, and metadata stop being ingested. In some cases thousands of endpoints go unhealthy simultaneously.
```
intxgo:
```suggestion
## Symptom

The Agent appears `ORPHANED` in Fleet, and Endpoint policy changes do not apply nor response actions can execute. In some cases thousands of endpoints go unhealthy simultaneously.

## Summary

Elastic Endpoint has lost communication with Elastic Agent, but is still protecting the system according to last known policy and sends all events and alerts to the stack.

## Common issues

### Elastic Agent not running

Elastic Agent crashed, has been explicitly disabled or removed.

### Communication between Agent and Endpoint services not working

Elastic Agent reports the Endpoint component as `FAILED` or `DEGRADED` with the message "Failed: endpoint service missed 3 check-ins". For troubleshooting see next symptom below.

## Symptom

Elastic Agent appears `UNHEALTHY` in fleet and Elastic Defend integration indicates errors. Elastic Agent reports the Endpoint component as `FAILED` or `DEGRADED` with the message "Failed: endpoint service missed 3 check-ins". Endpoint events, alerts, and metadata stop being ingested. In some cases thousands of endpoints go unhealthy simultaneously.
```

package/endpoint/docs/knowledge_base/missed_checkins/linux_missed_checkins.md
```diff
@@ -0,0 +1,102 @@
+---
+type: automatic_troubleshooting
+sub_type: missed_checkins
+os: [Linux]
+date: '2026-03-09'
+---
+
```
intxgo, line 7: Similar to windows case, please also explain Elastic Agent crash case.

package/endpoint/docs/knowledge_base/incompatible_software/incompatible_software_third_party.md
```diff
@@ -0,0 +1,90 @@
...
+---
+
+## Symptom
+
+A third-party application malfunctions, crashes, or causes system instability when Elastic Defend (or the Endgame Sensor) is installed. Common manifestations include: application crashes (e.g. browsers closing unexpectedly), network policies or VPN functionality breaking, CPU spikes rendering the system unresponsive, or virtual machines hanging during policy changes. The issue resolves when the endpoint security product is uninstalled or specific protection features are disabled.
```
intxgo:
```suggestion
A third-party application malfunctions, crashes, or causes system instability when Elastic Defend is installed. Common manifestations include: application crashes (e.g. browsers closing unexpectedly), network policies or VPN functionality breaking, CPU spikes rendering the system unresponsive, or virtual machines hanging during policy changes. The issue resolves when the endpoint security product is uninstalled or specific protection features are disabled.
```

why that discontinued product appeared here? Furthermore it kind of suggested that Endpoint and Sensor are recommended to run together.

package/endpoint/docs/knowledge_base/incompatible_software/incompatible_software_third_party.md
```diff
@@ -0,0 +1,90 @@
...
+
+
+## Summary
+
+Elastic Defend operates at multiple system levels — kernel-mode drivers for file system and network filtering, user-mode DLL injection for exploit protection (Endgame Sensor), and eBPF probes on Linux for host isolation and event collection. These deep integrations can conflict with other software that operates at similarly privileged levels: other security products with kernel drivers, eBPF-based networking tools, VPN clients that intercept DNS or network traffic, and applications that generate extreme volumes of system activity.
```
intxgo:
```suggestion
Elastic Defend operates at multiple system levels — kernel-mode drivers for file system and network filtering, user-mode DLL injection for exploit protection, and eBPF probes on Linux for host isolation and event collection. These deep integrations can conflict with other software that operates at similarly privileged levels: other security products with kernel drivers, eBPF-based networking tools, VPN clients that intercept DNS or network traffic, and applications that generate extreme volumes of system activity.
```

package/endpoint/docs/knowledge_base/high_cpu/windows_high_cpu.md
```diff
@@ -0,0 +1,96 @@
...
+
+## Common issues
+
+### Third-party security product conflicts
+
+When another security product (Silverfort, Symantec/Broadcom, N-Able AV Defender, CrowdStrike) runs alongside Elastic Defend, each product intercepts the other's file and process activity, creating a scanning feedback loop. This commonly manifests as sustained 100% CPU with both products' processes at the top of task manager.
+
+Silverfort is a particularly acute case on Domain Controllers because its `SilverfortServer.exe` generates over 10,000 TCP connections per minute via WinDivert, each producing a network event that Elastic Defend must process. Combined with Malicious Behavior Protection requiring network event enrichment for the rules engine, this can saturate CPU and eventually cause blue screens.
+
+To confirm, check `elastic-endpoint top` for the third-party product's processes. If they dominate `overall.week_ms` in the metrics, add them as **Trusted Applications** in Elastic Defend. Also add Elastic Defend's paths to the third-party product's exclusion list.
```
intxgo:
```suggestion
Add all 3rd party security applications as **Trusted Applications** in Elastic Defend to break feedback loops. Also add Elastic Defend's paths to the third-party product's exclusion list.
To confirm the problem on Endpoint side check `elastic-endpoint top` for the third-party product's processes. If they dominate `overall.week_ms` in the metrics. 
```

package/endpoint/docs/knowledge_base/bsod/windows_bsod_endpoint_driver.md
```diff
@@ -0,0 +1,80 @@
+---
+type: automatic_troubleshooting
+sub_type: bsod
+os: [Windows]
+date: '2026-03-09'
+---
+
+## Symptom
+
+A Windows system running Elastic Defend experiences a Blue Screen of Death (BSOD) or kernel crash. The memory dump analysis references `elastic_endpoint_driver.sys` or `elastic-endpoint-driver.sys`. Common bugcheck codes include `IRQL_NOT_LESS_OR_EQUAL`, `KERNEL_MODE_HEAP_CORRUPTION` (0x13A), `SYSTEM_SERVICE_EXCEPTION` (0x3B), or `PAGE_FAULT_IN_NONPAGED_AREA`. The crash may occur shortly after an agent upgrade, during heavy I/O workloads, or after the system has been running for some time with many network connections. In severe cases the system enters a boot loop.
```
intxgo:
```suggestion
A Windows system running Elastic Defend experiences a Blue Screen of Death (BSOD) or kernel crash. The memory dump analysis references `elastic_endpoint_driver.sys` or `elastic-endpoint-driver.sys`. The crash may occur shortly after an agent upgrade, 3rd party security product installation or it's configuration change, during heavy I/O workloads, or after the system has been running for some time with many network connections. In severe cases the system enters a boot loop.
```

I don't think we're commonly causing crashes, so this sentence makes absolutely no sense to me.

intxgo (commented):
Flushing some comments/suggestions

package/endpoint/docs/knowledge_base/incompatible_software/incompatible_software_third_party.md
ferullo:
```suggestion
Elastic Defend operates at multiple system levels — kernel-mode drivers for file system and network filtering on Windows, system extension and network content filtering on macOS, and eBPF probes on Linux. These deep integrations can conflict with other software that operates at similarly privileged levels: other security products with kernel drivers, eBPF-based networking tools, VPN clients that intercept DNS or network traffic, and applications that generate extreme volumes of system activity.
```

package/endpoint/docs/knowledge_base/incompatible_software/incompatible_software_third_party.md
```diff
@@ -0,0 +1,90 @@
...
+
+The root cause is a bug in the Endgame Sensor's hook manager linked list. When the maximum number of hooks (4) for a particular function in a module is reached, the error handling path frees a `HOOK_ENTRY` that is still part of the global hook list. The next traversal of the list encounters corrupted `flink`/`blink` pointers and crashes. Windows updates that add new DLLs to browser processes increase the number of hooks installed, making it more likely to hit the limit and trigger the bug.
+
+**Fixed in**: Endgame Sensor 3.65.3.
+
+**Workaround** (for older sensor versions): In the Endgame policy under Threats > Exploit, uncheck Chrome and Edge to disable exploit protection injection into those browsers. Additionally, under Settings > Event Collection > Windows Event Collection, disable beta event sources. This prevents `esensordbi.dll` from being injected into the affected processes.
+
+Disabling exploit protection for browsers reduces defense against browser exploitation. Keep browsers and the OS fully patched to minimize the window of exposure.
```
ferullo:
```suggestion
```

package/endpoint/docs/knowledge_base/incompatible_software/incompatible_software_third_party.md
```diff
@@ -0,0 +1,90 @@
...
+
+VPN clients such as Twingate and Palo Alto GlobalProtect can conflict with Elastic Defend's network event collection on macOS. The conflict typically manifests as DNS resolution failures — internal domains fail to resolve while the VPN is connected and Elastic Defend is running, but resolve correctly when either product is disabled.
+
+Elastic Defend's network monitoring on macOS intercepts DNS traffic to generate DNS events. VPN clients that implement split-tunnel DNS or custom DNS routing may conflict with this interception, causing DNS queries to be dropped or routed incorrectly.
+
+To diagnose: compare DNS resolution behavior with Elastic Defend enabled vs disabled while the VPN is connected. Use `nslookup` or `dig` to test resolution of both internal (VPN-routed) and external domains.
+
+To resolve: if DNS events are not essential, disable DNS event collection in the Elastic Defend policy. If DNS events are needed, add the VPN client process as a Trusted Application to stop Endpoint from monitoring its network activity. If the conflict persists, set `mac.advanced.capture.env.dns: false` in advanced policy settings to disable DNS event capture at a lower level.
```
ferullo: `mac.advanced.capture.env.dns` does not exist. I don't see any advanced option that can do this, @ricardo-estc is there one?

package/endpoint/docs/knowledge_base/incompatible_software/incompatible_software_third_party.md
```diff
@@ -0,0 +1,90 @@
...
+
+### Microsoft DFSR replication issues with file monitoring
+
+On Windows Server systems running Distributed File System Replication (DFSR), Elastic Defend's file system filter driver can interfere with replication operations. Symptoms include replication backlogs, unexpected staging folder growth, or DFSR service errors in the event log. The issue is exacerbated when rollback self-healing (`windows.advanced.artifacts.global.rollback.self_healing.enabled`) is active, as Elastic Defend may detect and attempt to roll back legitimate DFSR staging operations that involve suspicious file patterns.
```
ferullo:
```suggestion
On Windows Server systems running Distributed File System Replication (DFSR), Elastic Defend's file system filter driver can interfere with replication operations. Symptoms include replication backlogs, unexpected staging folder growth, or DFSR service errors in the event log. The issue is exacerbated when rollback self-healing (`windows.advanced.alerts.rollback.self_healing.enabled`) is active, as Elastic Defend may detect and attempt to roll back legitimate DFSR staging operations that involve suspicious file patterns.
```

package/endpoint/docs/knowledge_base/incompatible_software/incompatible_software_third_party.md
```diff
@@ -0,0 +1,90 @@
...
+
+
+## Investigation priorities
+
+1) Identify the conflicting third-party software by correlating the onset of symptoms with software installation, updates, or configuration changes. Query `logs-endpoint.events.process-*` for recently started processes that match known conflicting software.
+2) Determine which Elastic Defend subsystem is involved — kernel network driver (`windows.advanced.kernel.network`), eBPF probes (`linux.advanced.host_isolation.allowed`), exploit protection DLL injection (Endgame policy), or file system filter driver. Each has different mitigation controls.
+3) Check `metrics-endpoint.metrics-*` for CPU usage patterns and correlate with `logs-endpoint.events.process-*` for processes generating high event volumes. Use `elastic-endpoint top` on the affected host to identify which processes and event types drive resource consumption.
+4) For browser crashes on Windows, collect crash dumps from `%HOMEPATH%\AppData\Local\Google\Chrome\User Data\Crashpad\reports` or `%LOCALAPPDATA%\CrashDumps` and check for `esensordbi.dll` in the faulting module. Verify the Endgame Sensor version and check if a fix is available.
```
ferullo:
```suggestion
4) For browser crashes on Windows, collect crash dumps from `%HOMEPATH%\AppData\Local\Google\Chrome\User Data\Crashpad\reports` or `%LOCALAPPDATA%\CrashDumps`.
```

package/endpoint/docs/knowledge_base/missed_checkins/linux_missed_checkins.md
```diff
@@ -0,0 +1,102 @@
...
+```
+sudo semanage fcontext -a -t bin_t '/opt/Elastic/Endpoint(/.*)?'
+sudo restorecon -Rv /opt/Elastic/Endpoint
+sudo systemctl restart ElasticEndpoint.service
+```
+
+This sets the correct SELinux file context so systemd is permitted to execute the Endpoint binary. The fix persists across Endpoint upgrades as long as the `semanage` rule remains.
+
+### "No valid comms client available" in Endpoint logs
```
ferullo: This log message is wrong. @intxgo do you know what log is correct here?

package/endpoint/docs/knowledge_base/missed_checkins/linux_missed_checkins.md
```diff
@@ -0,0 +1,102 @@
...
+
+This message in the Endpoint state logs (`endpoint-*.log`) indicates that the Endpoint process started but cannot establish a communication channel with Elastic Agent. Common causes:
```
ferullo:
```suggestion
This message in the Endpoint state logs (`logs-elastic_agent.endpoint_security-*`) indicates that the Endpoint process started but cannot establish a communication channel with Elastic Agent. Common causes:
```

package/endpoint/docs/knowledge_base/missed_checkins/linux_missed_checkins.md
```diff
@@ -0,0 +1,102 @@
...
+
+- The Agent process restarted or is not yet ready to accept connections when Endpoint initializes.
+- A firewall rule or security policy is blocking localhost connections on ports 6788/6789.
+- The Endpoint process is aborting shortly after startup due to another issue (look for `Aborting due to signal` immediately after the comms error).
```
ferullo, line 46:
```suggestion
- The Endpoint process is aborting shortly after startup due to another issue (look for `Aborting due to signal` immediately after the comms error).
- Another system process is listening on these ports.
```

package/endpoint/docs/knowledge_base/missed_checkins/linux_missed_checkins.md
```diff
@@ -0,0 +1,102 @@
...
+- Crash messages in `journalctl -u ElasticEndpoint.service`
+- `Aborting due to signal` messages in Endpoint logs at `logs-elastic_agent.endpoint_security-*`
+
+If the crash is reproducible, collect the core dump and Endpoint diagnostic output (`/opt/Elastic/Endpoint/elastic-endpoint diagnostics`) and report to Elastic support.
+
+### Port conflict on 6788/6789
+
+Elastic Agent and Endpoint communicate over TCP ports 6788 and 6789 on localhost. If another process is already bound to one of these ports, Endpoint cannot establish its communication channel with Agent and will miss check-ins.
+
+To check: run `ss -tlnp | grep -E '6788|6789'` on the affected host. If a non-Elastic process is listening on either port, it must be reconfigured to use a different port, or the Endpoint communication port can be changed via advanced policy settings.
```
ferullo: > or the Endpoint communication port can be changed via advanced policy settings.

Is that true?

package/endpoint/docs/knowledge_base/missed_checkins/linux_missed_checkins.md
```diff
@@ -0,0 +1,102 @@
...
+
+## Investigation priorities
+
+1) Check `journalctl -u ElasticEndpoint.service` for startup failures, exit codes (203, 231), and crash messages
+2) Check `logs-elastic_agent.endpoint_security-*` for error messages: look for `No valid comms client available`, `Aborting due to signal`, `exit status 203`, or `exit status 231`
+3) Check SELinux audit logs with `ausearch -m avc -c '(endpoint)'` for execution denials
+4) Check `.fleet-agents*` for the agent's `last_checkin` timestamp and current status to confirm the endpoint is actually missing check-ins vs. the entire agent being offline
+5) Run `ss -tlnp | grep -E '6788|6789'` on the host to check for port conflicts
+6) Check `dmesg` and `journalctl -k` for OOM killer events targeting the Endpoint process
+7) Check `metrics-endpoint.policy-*` for FAILED policy responses indicating the endpoint tried to start but could not apply its policy
```
ferullo:
```suggestion
```

package/endpoint/docs/knowledge_base/missed_checkins/windows_missed_checkins.md
```diff
@@ -0,0 +1,88 @@
...
+
+This is a symptom of the same shutdown race condition described above. The partial uninstall corrupts the Endpoint's code signing state. The self-healing logic introduced in 8.4.3 attempts to repair on reboot, but the repair can itself fail on certain OS versions if the rollback zip file was created but the reinstall was interrupted.
+
+To fix: follow the same remediation as for exit status 231 above. Upgrade to Agent 8.8.1+ to prevent the triggering race condition.
+
+### Unicode conversion crash on non-English Windows (pre-8.13.3)
+
+On Windows systems using non-English locales (confirmed on Chinese GB2312 and Korean editions), a unicode conversion bug causes the Endpoint process to crash in a loop immediately after startup. The crash occurs during kernel communication initialization and produces hundreds of restarts per log file. Agent reports the Endpoint as FAILED because it never completes startup.
+
+The bug is tracked in elastic/security#1973 and was fixed in Endpoint 8.13.3. Check the Endpoint logs for crash stack traces mentioning `CompletionPort.cpp` or `BigChief.cpp` during the `Connecting to minifilter` phase.
```
ferullo:
```suggestion
```

package/endpoint/docs/knowledge_base/output_config/output_kafka_message_size.md
```diff
@@ -0,0 +1,95 @@
...
+
+
+## Common issues
+
+### Kafka `Message size too large` broker rejection
+
+The Kafka broker rejects messages that exceed its configured `max.message.bytes` limit. The endpoint logs the error:
+
+```
+KafkaClient.cpp:561 KafkaClient failed to deliver record with unrecoverable error: Broker: Message size too large [10] | non-retriable
```
ferullo: Drop filename/line number since that could easily change.

```suggestion
KafkaClient failed to deliver record with unrecoverable error: Broker: Message size too large [10] | non-retriable
```

package/endpoint/docs/knowledge_base/output_config/output_kafka_message_size.md
```diff
@@ -0,0 +1,95 @@
...
+To confirm whether messages are being dropped (rather than the error no longer occurring): if the `Message size too large` error no longer appears in `logs-elastic_agent.endpoint_security-*`, it means the condition is not being triggered — not that messages are being silently dropped. Dropped messages still produce the error log entry on the first attempt.
+
+### Logstash output disconnection causing CPU spike
+
+When the Logstash output becomes unreachable, the endpoint enters a reconnection loop. On versions 8.8.0 through 8.13.3, a bug in the retry logic caused the `elastic-endpoint` process to consume 100% of one CPU core within minutes of losing connectivity. The CPU returns to normal within approximately 40 seconds of connectivity being restored.
+
+Characteristic log pattern during the outage:
+
+```
+AgentContext.cpp:565 Endpoint is setting status to DEGRADED, reason: Unable to connect to output server
```
ferullo:
```suggestion
Endpoint is setting status to DEGRADED, reason: Unable to connect to output server
```

package/endpoint/docs/knowledge_base/output_config/output_kafka_message_size.md
```diff
@@ -0,0 +1,95 @@
...
+LogstashClient.cpp:662 SSL handshake with Logstash server at [host]:[port] encountered an error
```
ferullo:
```suggestion
SSL handshake with Logstash server at [host]:[port] encountered an error
```

package/endpoint/docs/knowledge_base/output_config/output_kafka_message_size.md
```diff
@@ -0,0 +1,95 @@
...
+BulkQueueConsumer.cpp:192 Logstash connection is down
```
ferullo:
```suggestion
Logstash connection is down
```

package/endpoint/docs/knowledge_base/output_config/output_kafka_message_size.md
```diff
@@ -0,0 +1,95 @@
...
+```
+
+These messages repeat every 20 seconds while the output is unreachable. The CPU spike is caused by the endpoint's C++ Logstash client, not by the Beats/Agent Go-based outputs — removing the Endpoint Security integration from the policy immediately drops CPU back to normal even while Logstash remains down, confirming the issue is specific to Elastic Defend's output implementation.
+
+**Fixed in 8.13.4** (endpoint-dev PR #13065). On fixed versions, the endpoint still logs connection errors but uses a proper exponential backoff that does not spin CPU.
```
ferullo:
```suggestion
**Fixed in 8.13.4**. On fixed versions, the endpoint still logs connection errors but uses a proper exponential backoff that does not spin CPU.
```

package/endpoint/docs/knowledge_base/output_config/output_kafka_message_size.md
```diff
@@ -0,0 +1,95 @@
...
+
+### Logstash output event backlog after reconnection
+
+After a prolonged Logstash disconnection, the endpoint accumulates unsent events in its internal queue. When connectivity is restored, the endpoint resumes sending but starts with the oldest queued events. The delivery delay equals the duration of the outage — if the output was down for one hour, events will be approximately one hour behind real time until the backlog is drained.
+
+The backlog does not self-resolve faster than real-time ingestion because the endpoint sends queued events at the same rate as live events. On deployments with high event volume (e.g. 45,000+ agents behind a Logstash cluster), this delay can persist for the same duration as the outage.
```
ferullo, line 67: @brian-mckinney is this or was it ever true?

package/endpoint/docs/knowledge_base/output_config/output_kafka_message_size.md
```diff
@@ -0,0 +1,95 @@
...
+
+The endpoint implements a backoff algorithm for reconnection attempts. If the backoff interval has grown large due to repeated failures, the connection may not be reattempted immediately even after connectivity is restored. Running `elastic-endpoint test output` on the affected host manually cancels the backoff and triggers an immediate reconnection attempt. The corresponding log entry is:
+
+```
+ServiceCommsFunctions.cpp:234 Canceling output backoff due to 'test output' command
```
ferullo:
```suggestion
Canceling output backoff due to 'test output' command
```

package/endpoint/docs/knowledge_base/trusted_apps/trusted_apps_not_effective.md
ferullo: > - implicitly meaning it only works on executables

This is true for standard trusted apps but Advanced Trusted Apps lets users trust non-process activity. The original version seems more correct to me.

package/endpoint/docs/knowledge_base/trusted_apps/trusted_apps_not_effective.md
ferullo: Replace `{{elastic-defend}}` with Elastic Defend.

> we don't have a commonly used feature "Suppress protections (trust) everything matching the path: xxx"

It's not obvious but the solution are Endpoint Alert Exceptions for that (either based on `file.path MATCHES c:\whatever\*` or `process.executable MATCHES c:\whatever\*`

package/endpoint/docs/knowledge_base/trusted_apps/trusted_apps_not_effective.md
```diff
@@ -0,0 +1,77 @@
...
+Common scenarios where an Endpoint Alert Exception should be used instead:
+- Malware engine false positive on a known-good binary: add an exception on the file hash.
+- Malware On Write blocking a legitimate file: Trusted Applications operate at the process level and do not prevent Malware On Write from scanning files as they are written to disk by another process (e.g. `explorer.exe` extracting an archive). Use an Endpoint Alert Exception for the specific file.
+- Process injection alerts for legitimate software: use an Endpoint Alert Exception targeting the specific alert fields.
+
+### Trusted Application assigned to wrong policy or OS
+
+A Trusted Application entry is scoped to specific integration policies. If the entry is not assigned to the policy covering the affected endpoint, it has no effect. Similarly, a Trusted Application configured for the wrong OS will not apply.
+
+In the Kibana Trusted Applications UI, verify the "Assignment" column shows the correct integration policy. If using "Global" assignment, confirm the affected endpoint's policy is not explicitly excluded.
```
ferullo: > If using "Global" assignment, confirm the affected endpoint's policy is not explicitly excluded.

Is that a feature I'm not aware of?

package/endpoint/docs/knowledge_base/trusted_apps/trusted_apps_not_effective.md
intxgo: agree, we need to walk-by-hand the users here, so explain them TA is not what they think is the right tool -> point them towards exceptions... and here is another rabbit hole. I've recently tried to ask public LLM how to make an exception in Elastic Defend for a particular alert, it got it all wrong! LLMs (likewise me, looking at the public documents) are really confused about how to set Endpoint exceptions. The whole documentation is written for BEATS, loosely adding in Elastic Defend to the picture as "Endpoint exception" but IMO this is really confusing when you're having actions/processes blocked on your machine, going to Kibana to "suppress them". The verbiage in the UI and docs could be much more improved. I know I'm biased as Endpoint developer, but the UI/doc kind of hide/obscure Endpoint alerts and exception lists mechanics.

package/endpoint/docs/knowledge_base/trusted_apps/trusted_apps_not_effective.md
intxgo: @ferullo I have to verify that, but I seriously doubt it, last time I walked the code, TA was evaluated only on execution callbacks, so anything like file access/modification/creation won't work here, but that's what people assume this can do

intxgo:
A gneral question, is it going to be stack versioned? Is the current PR for the lastest 9.3 stack, or what? I agree with other comments here that it doesn't make sense to train AI on past already fixed bug symptoms, also quoting "Endgame" or "sensor" really makes no sense here, it's unrelated product, even not supported anymore.

package/endpoint/docs/knowledge_base/missed_checkins/linux_missed_checkins.md
intxgo: I'll park it for tomorrow, thanks

package/endpoint/docs/knowledge_base/trusted_apps/trusted_apps_not_effective.md
joeypoon: yeah, this is an optional field

package/endpoint/docs/knowledge_base/missed_checkins/linux_missed_checkins.md
intxgo: Indeed, that message is for output only.

Endpoint -> Agent connection issue can be breaking in multiple places, I think we can point users at the 2 major logs

```diff
- ### "No valid comms client available" in Endpoint logs
+ ### "Unable to retrieve connection info from Agent" or "Agent GRPC connection failed to establish within deadline" in Endpoint logs

File: package/endpoint/docs/knowledge_base/missed_checkins/linux_missed_checkins.md (hunk 40 above)
Comment (intxgo) [reply package/endpoint/docs/knowledge_base/missed_checkins/linux_missed_checkins.md]: old Endpoints hardcoded 6788, newer endpoints hardcode named pipe name

6789 was always given by Agent as bootstrap message, if it can be configured on Agent, then Endpoint will use what's given in bootstrap

**Comment by ferullo** [comment]
> A gneral question, is it going to be stack versioned? Is the current PR for the lastest 9.3 stack, or what? I agree with other comments here that it doesn't make sense to train AI on past already fixed bug symptoms, also quoting "Endgame" or "sensor" really makes no sense here, it's unrelated product, even not supported anymore.

Since this is in the integration package it can update at any time. In practice we'll likely update it with the latest Stack and only backport important knowledge to older Stacks.

Current 9.x stacks support all Agents and Endpoints going back to 8.19, so knowledge of all nuances/issues/problems with Agent 8.19->current are relevant. Stack-side problems older than the current Stack version are not relevant.

Yes, all references to Endgame should be removed.

File: package/endpoint/docs/knowledge_base/bsod/windows_bsod_endpoint_driver.md [hunk 51]
```diff
@@ -0,0 +1,69 @@
...
+## Symptom
+
+A Windows system running Elastic Defend experiences a Blue Screen of Death (BSOD) or kernel crash. The memory dump analysis references `elastic_endpoint_driver.sys` or `elastic-endpoint-driver.sys`. The crash may occur shortly after an agent upgrade, 3rd party security product installation or its configuration change, during heavy I/O workloads, or after the system has been running for some time with many network connections. In severe cases the system enters a boot loop.
+
+
+## Summary
+
+Elastic Defend uses a kernel-mode driver (`elastic_endpoint_driver.sys`) for file system filtering, network monitoring, and process/object callbacks. Most BSOD issues traced to the endpoint driver fall into a few categories: regressions introduced in specific driver versions, conflicts with other kernel-mode drivers (third-party security products), or running on unsupported OS versions.
+
+Collecting a full kernel memory dump and sharing it with Elastic Support is essential for root-cause determination. The bugcheck code alone is not sufficient — the faulting call stack identifies which code path triggered the crash.
```
ferullo:
```suggestion
Collecting a full kernel memory dump and sharing it with Elastic Support is essential for root-cause determination. The bugcheck code alone is not sufficient — the faulting call stack identifies which code path triggered the crash. Just because Elastic Defend is in the calls stack does not mean it is responsible for the crash.
```

package/endpoint/docs/knowledge_base/endpoint_exceptions/false_positive_malware_detection.md
```diff
@@ -0,0 +1,73 @@
...
+- **Blocklist**: A separate artifact that prevents known-malicious binaries from executing. It is NOT the inverse of an exception — blocklist entries block execution, while exceptions suppress alerts. If a process matches the Global Exception List (e.g. signed by a trusted vendor like Microsoft), the User Blocklist is not evaluated, which can cause blocklist entries to appear ineffective for certain trusted-signed binaries.
+
+Key distinction: An **Endpoint Alert Exception** suppresses alerts while keeping monitoring active. A **Trusted Application** stops Endpoint from monitoring the process entirely. For false positives, an Endpoint Alert Exception is almost always the correct choice.
+
+
+## Common issues
+
+### Malware engine false positive on legitimate software
+
+The malware ML model assigns a score to a file's binary content. Legitimate software can occasionally score above the threshold (default 0.58), especially newly released or niche software that the model has not been trained on. The alert's `file.Ext.malware_classification.score` and `file.Ext.malware_classification.threshold` fields show the model's confidence.
```
ferullo: The exact value changes every time the model is changed

```suggestion
The malware ML model assigns a score to a file's binary content. Legitimate software can occasionally score above the threshold, especially newly released or niche software that the model has not been trained on. The alert's `file.Ext.malware_classification.score` and `file.Ext.malware_classification.threshold` fields show the model's confidence.
```

package/endpoint/docs/knowledge_base/high_cpu/linux_high_cpu.md
```diff
@@ -0,0 +1,102 @@
...
+
+## Common issues
+
+### Monitoring scripts spawning many short-lived child processes
+
+Monitoring and automation scripts that run on a schedule (cron, systemd timers) and spawn many child processes are the most common cause of high CPU on Linux. A single monitoring script invoking `curl`, `mysql`, `ssh`, `grep`, `sed`, `awk`, and `bash` in rapid succession generates a burst of process creation events, each of which Elastic Defend must enrich and evaluate against behavioral rules.
+
+A typical pattern is hourly CPU spikes lasting 5–10 minutes, aligning with cron schedules (e.g. xx:31–xx:41 every hour).
+
+Adding the parent script as a Trusted Application stops monitoring of its process tree but does not prevent behavioral rules from firing if the rule matches on child process characteristics. On versions prior to 9.2, behavioral detections still fire for trusted processes. On 9.2+, behavioral detections are disabled for Trusted Applications.
```
ferullo:
```suggestion
Adding the parent script as a Trusted Application Descendants stops monitoring of its process tree but does not prevent behavioral rules from firing if the rule matches on child process characteristics. On versions prior to 9.2, behavioral detections still fire for trusted processes. On 9.2+, behavioral detections are disabled for Trusted Applications.
```

package/endpoint/docs/knowledge_base/incompatible_software/incompatible_software_third_party.md
```diff
@@ -0,0 +1,79 @@
...
+
+The issue occurs because the kernel network driver's initialization or reconfiguration during a policy change blocks other system operations in the VM's constrained environment. Disabling network event collection in the policy prevents the hang.
+
+If network visibility is required in VDI environments, test policy changes during maintenance windows and consider using `windows.advanced.kernel.network: false` as an alternative to full network event disablement — this disables the kernel driver while allowing behavioral protections that do not depend on kernel-level network capture to remain active.
+
+
+## Investigation priorities
+
+1) Identify the conflicting third-party software by correlating the onset of symptoms with software installation, updates, or configuration changes. Query `logs-endpoint.events.process-*` for recently started processes that match known conflicting software.
+2) Determine which Elastic Defend subsystem is involved — kernel network driver (`windows.advanced.kernel.network`), eBPF probes (`linux.advanced.host_isolation.allowed`), exploit protection DLL injection (Endgame policy), or file system filter driver. Each has different mitigation controls.
```
ferullo:
```suggestion
2) Determine which Elastic Defend subsystem is involved — kernel network driver (`windows.advanced.kernel.network`), eBPF probes (`linux.advanced.host_isolation.allowed`), or file system filter driver. Each has different mitigation controls.
```

package/endpoint/docs/knowledge_base/missed_checkins/windows_missed_checkins.md
```diff
@@ -0,0 +1,105 @@
...
+
+## Common issues
+
+### Endpoint process crash (dump file generated)
+
+The Endpoint process can crash due to memory corruption, driver conflicts, or product bugs. When this happens, a crash dump file (`elasticendpoint.dmp`) is written to `C:\Program Files\Elastic\Endpoint\` or `C:\Program Files\Elastic\Endpoint\cache\CrashDumps\`. After a crash, the Endpoint service attempts to restart automatically, but if the crash is persistent (e.g. corrupted binary in memory), it can crash repeatedly — up to 15-16 times in quick succession — before Agent marks it as FAILED.
+
+To investigate: check for `.dmp` files at the paths above. If a dump exists, run WinDbg with `!analyze -v` against it to identify the crash location. Check `logs-elastic_agent.endpoint_security-*` for patterns like `Aborting due to signal` or unhandled exception stack traces.
+
+If the crash is caused by memory corruption (corrupted `.text` section of the Endpoint binary, random bit flips), it may be a transient hardware issue. Run Windows Memory Diagnostic or memtest86 to check for bad DIMMs. If the crash recurs at the same code location, collect the dump and report to Elastic support.
```
ferullo:
```suggestion
To investigate: check for `.dmp` files at the paths above. If a dump exists, collect it and report to Elastic support.
```

package/endpoint/docs/knowledge_base/missed_checkins/windows_missed_checkins.md
```diff
@@ -0,0 +1,105 @@
...
+
+Elastic Agent and Endpoint communicate over TCP ports 6788 and 6789 on localhost. If another process is already bound to one of these ports, Endpoint cannot establish its communication channel with Agent and will miss check-ins.
+
+To check: run `netstat -ano | findstr "6788 6789"` on the affected host. If a non-Elastic process is listening on either port, it must be reconfigured to use a different port.
+
+### Artifact download or verification failure causing mass unhealthiness
+
+When Endpoint receives a new policy, it must download and verify user artifact manifests (trusted apps, exception lists, blocklists) from Fleet Server. If the artifact manifest is corrupted — for example, `compression_algorithm` is set to `none` instead of `zlib` due to a known stack bug fixed in 8.7.1 — Endpoint rejects all artifacts and the policy fails to apply. Logs show `All artifacts are being rejected because endpoint-trustlist-windows-v1 is invalid` and `Failed to process artifact manifest`.
+
+At scale, simultaneous policy pushes can also overwhelm Fleet Server, producing HTTP 429 responses and causing a cascading failure where all endpoints go unhealthy at once.
```
ferullo:
```suggestion
At scale, simultaneous policy pushes can also overwhelm Fleet Server, producing HTTP 429 responses and causing a cascading failure where all endpoints go unhealthy at once. This issue with resolve itself over time.
```

package/endpoint/docs/knowledge_base/high_cpu/linux_high_cpu.md
```diff
@@ -0,0 +1,102 @@
...
+---
+
+## Symptom
+
+Elastic Defend's `elastic-endpoint` process consumes sustained high CPU on Linux hosts, causing system slowdown, degraded application performance, or resource exhaustion. CPU may spike during specific intervals (cron jobs, monitoring scripts), during policy application, or continuously due to output connectivity failures.
+
+
+## Summary
+
+Elastic Defend on Linux uses eBPF or tracefs to monitor process, file, network, and DNS activity in real time. Each event is enriched, hashed, evaluated against behavioral rules, and forwarded to the configured output. The most common drivers of high CPU on Linux are monitoring scripts that spawn many short-lived child processes (each generating process events that trigger behavioral rules), output server disconnections causing retry storms, the Events plugin hashing large binaries during policy application with an empty cache, and memory scanning of large processes.
```
ferullo:
```suggestion
Elastic Defend on Linux uses eBPF or tracefs to monitor process, file, network, and DNS activity in real time. Each event is enriched, hashed, evaluated against behavioral rules, and forwarded to the configured output. The most common drivers of high CPU on Linux are monitoring scripts that spawn many short-lived child processes (each generating process events that trigger behavioral rules), output server disconnections causing retry storms, hashing large binaries during policy application with an empty cache, and memory scanning of large processes.
```

package/endpoint/docs/knowledge_base/high_cpu/linux_high_cpu.md
```diff
@@ -0,0 +1,102 @@
...
+- `SSL handshake with Logstash server at [host]:[port] encountered an error`
+- `Logstash connection is down`
+
+CPU returns to normal within approximately 40 seconds after connectivity is restored. The command `sudo /opt/Elastic/Endpoint/elastic-endpoint test output` can be used to verify output connectivity.
+
+**Fixed in 8.13.4**. On fixed versions, the endpoint still logs connection errors but uses a proper exponential backoff that does not spin CPU. Check `logs-elastic_agent.endpoint_security-*` for the error patterns above.
+
+For Kafka outputs, `Message size too large` errors cause repeated delivery failures. On 8.18.3+, oversized messages are dropped gracefully instead of retried indefinitely.
+
+### Events plugin hung hashing large binaries during policy application
```
ferullo:
```suggestion
### A hang when hashing large binaries during policy application
```

package/endpoint/docs/knowledge_base/high_cpu/linux_high_cpu.md
```diff
@@ -0,0 +1,102 @@
...
+
+During policy application, Elastic Defend hashes all running processes and their executables. When the file cache (`/opt/Elastic/Endpoint/state/cache.db`) is empty — first install, cache deleted, or after upgrade — the endpoint must hash every binary from scratch. Large binaries (e.g. Oracle) can cause the hashing to drive CPU to 100% for extended periods.
+
+The endpoint will report status as CONFIGURING during this time:
+- `Endpoint is setting status to CONFIGURING, reason: Policy Application Status`
+
+On first run with an empty cache, the CONFIGURING phase can take 5–30 minutes depending on the number and size of running processes. This is expected behavior. Subsequent restarts are fast because the cache persists.
+
+This behavior was improved in 8.14+ where the hashing is less likely to hang the Events plugin. On older versions, restarting the service (`sudo systemctl restart elastic-endpoint`) may temporarily resolve the hang.
```
ferullo:
```suggestion
This behavior was improved in 8.14+ where the hashing is less likely to hang. On older versions, restarting the service (`sudo systemctl restart elastic-endpoint`) may temporarily resolve the hang.
```

package/endpoint/docs/knowledge_base/high_cpu/windows_high_cpu.md
```diff
@@ -0,0 +1,95 @@
...
+Use `elastic-endpoint top` on the affected host to identify which processes and internal processing stages (MLWR, BHVR, RULES, etc.) consume the most CPU. Query `metrics-endpoint.metrics-*` for `Endpoint.metrics.system_impact` to identify the top processes by `overall.week_ms` remotely.
+
+
+## Common issues
+
+### Third-party security product conflicts
+
+When another security product (Silverfort, Symantec/Broadcom, N-Able AV Defender, CrowdStrike) runs alongside Elastic Defend, each product intercepts the other's file and process activity, creating a scanning feedback loop. This commonly manifests as sustained 100% CPU with both products' processes at the top of task manager.
+
+Silverfort is a particularly acute case on Domain Controllers because its `SilverfortServer.exe` generates over 10,000 TCP connections per minute via WinDivert, each producing a network event that Elastic Defend must process. Combined with Malicious Behavior Protection requiring network event enrichment for the rules engine, this can saturate CPU and eventually cause blue screens.
```
ferullo:
```suggestion
Silverfort is a particularly acute case on Domain Controllers because its `SilverfortServer.exe` generates over 10,000 TCP connections per minute via WinDivert, each producing a network event that Elastic Defend must process. Combined with Malicious Behavior Protection requiring network event enrichment for, this can saturate CPU and eventually cause blue screens.
```

package/endpoint/docs/knowledge_base/high_cpu/windows_high_cpu.md
```diff
@@ -0,0 +1,95 @@
...
+
+Add all 3rd party security applications as **Trusted Applications** in Elastic Defend to break feedback loops. Also add Elastic Defend's paths to the third-party product's exclusion list.
+To confirm the problem on Endpoint side check `elastic-endpoint top` for the third-party product's processes. If they dominate `overall.week_ms` in the metrics.
+
+For Silverfort specifically, if Trusted Applications are not sufficient due to sheer network event volume, set `windows.advanced.kernel.network` to `false` in the policy's advanced settings to stop network event generation at the kernel level. This disables host isolation capability.
+
+### High security event volume (logon/logoff events)
+
+On hosts with heavy authentication activity (print servers, file servers, Domain Controllers), Windows Security event log entries 4624 (Logon) and 4634 (Logoff) can generate tens of events per second. When Malicious Behavior Protection is enabled, Elastic Defend collects and processes Security events regardless of whether Security event collection is enabled in the policy, because the behavioral rules engine needs this data.
```
ferullo:
```suggestion
On hosts with heavy authentication activity (print servers, file servers, Domain Controllers), Windows Security event log entries 4624 (Logon) and 4634 (Logoff) can generate tens of events per second. When Malicious Behavior Protection is enabled, Elastic Defend collects and processes Security events regardless of whether Security event collection is enabled in the policy, because the malicious behavior rules need this data.
```

package/endpoint/docs/knowledge_base/high_cpu/windows_high_cpu.md
```diff
@@ -0,0 +1,95 @@
...
+
+A host generating 60+ logon/logoff events per second will see sustained high CPU from the rules engine processing pipeline. Query `metrics-endpoint.metrics-*` and check `Endpoint.metrics.system_impact` — if `authentication_events` dominates, this is the cause.
```
ferullo:
```suggestion
A host generating 60+ logon/logoff events per second will see sustained high CPU from the event or Malicious Behavior processing pipelines. Query `metrics-endpoint.metrics-*` and check `Endpoint.metrics.system_impact` — if `authentication_events` dominates, this is the cause.
```

ferullo (commented):
Looks really solid. I added just a few change requests.

package/endpoint/docs/knowledge_base/high_cpu/windows_high_cpu.md
```diff
@@ -0,0 +1,95 @@
...
+Use `elastic-endpoint top` on the affected host to identify which processes and internal processing stages (MLWR, BHVR, RULES, etc.) consume the most CPU. Query `metrics-endpoint.metrics-*` for `Endpoint.metrics.system_impact` to identify the top processes by `overall.week_ms` remotely.
+
+
+## Common issues
+
+### Third-party security product conflicts
+
+When another security product (Silverfort, Symantec/Broadcom, N-Able AV Defender, CrowdStrike) runs alongside Elastic Defend, each product intercepts the other's file and process activity, creating a scanning feedback loop. This commonly manifests as sustained 100% CPU with both products' processes at the top of task manager.
+
+Silverfort is a particularly acute case on Domain Controllers because its `SilverfortServer.exe` generates over 10,000 TCP connections per minute via WinDivert, each producing a network event that Elastic Defend must process. Combined with Malicious Behavior Protection requiring network event enrichment for, this can saturate CPU and eventually cause blue screens.
```
ferullo:
```suggestion
Silverfort is a particularly acute case on Domain Controllers because its `SilverfortServer.exe` generates over 10,000 TCP connections per minute via WinDivert, each producing a network event that Elastic Defend must process. Combined with Malicious Behavior Protection requiring network event enrichment, this can saturate CPU and eventually cause blue screens.
```

ferullo (approved):
LGTM (one very minor suggestion).

PR #730, opened 2026-03-18
Remove sample event strings that can trigger false positive alerts from other AV

ferullo:
## Change Summary
Change the sample api event so it doesn't trigger false positive alerts from other Antivirus.

## Release Target
* 9.4.0
* Will also backport to 8.19


pzl (approved):
the CI failures are from the added newline at end of file

bjmcnic (approved):

joe-desimone (approved):

PR #744, opened 2026-06-18
Add process.ai_agent.* custom documentation for Linux process events

stanek-michal:
Register process.ai_agent.name and process.ai_agent.is_descendant in all Linux process event documentation templates that already carry the analogous process.Ext.trusted_descendant field (fork/exec/exit, already_running, gid_change, memfd_create, ptrace, shmget, uid_change).

### Sample values

Sample document:

```json


```

### For mapping changes:

- [ ] I ran `make` after making the schema changes, and committed all changes
- [ ] If these field(s) are "exception"-able, I made a companion PR to Kibana adding it (see [Readme](https://github.com/elastic/endpoint-package#exceptionable))
- [ ] If this is a `metadata` change, I also updated both transform destination schemas to match

### For Transform changes:

- [ ] The new transform successfully starts in Kibana
- [ ] The corresponding transform destination schema was updated if necessary

PR #752, opened 2026-07-03
Add GenAI process fields from ECS

stanek-michal:
process.ai_agent.* was documented on Linux process events in #744, but the underlying field/schema definitions were never merged to main - here they are

(cherry picked from commit f645c16613a9bda71106f51443a1aab61defe002)

### Sample values

Sample document:

```json


```

### For mapping changes:

### For Transform changes:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,8 +1,8 @@
 ROOT_DIR := $(shell dirname $(realpath $(firstword $(MAKEFILE_LIST))))
 # we are intentionally pinning the ECS version here, when ecs releases a new version
 # we'll discuss whether we need to release a new package and bump the version here
-# 43a1a61a4a4db88e2de60da9019733610717ff7e is v8.10.0
-ECS_GIT_REF ?= 43a1a61a4a4db88e2de60da9019733610717ff7e
+# f04bb41e4fcc35befcaa94a1d9575bc61944520c adds process.ai_agent.* alpha fields.
+ECS_GIT_REF ?= f04bb41e4fcc35befcaa94a1d9575bc61944520c
 
 
 # This variable specifies to location of the package-storage repo. It is used for automatically creating a PR
--- a/Makefile
+++ b/Makefile
@@ -1,8 +1,8 @@
 ROOT_DIR := $(shell dirname $(realpath $(firstword $(MAKEFILE_LIST))))
 # we are intentionally pinning the ECS version here, when ecs releases a new version
 # we'll discuss whether we need to release a new package and bump the version here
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -1,8 +1,8 @@
 ROOT_DIR := $(shell dirname $(realpath $(firstword $(MAKEFILE_LIST))))
 # we are intentionally pinning the ECS version here, when ecs releases a new version
 # we'll discuss whether we need to release a new package and bump the version here
-# 43a1a61a4a4db88e2de60da9019733610717ff7e is v8.10.0
-ECS_GIT_REF ?= 43a1a61a4a4db88e2de60da9019733610717ff7e
+# f04bb41e4fcc35befcaa94a1d9575bc61944520c adds process.ai_agent.* alpha fields.
+ECS_GIT_REF ?= f04bb41e4fcc35befcaa94a1d9575bc61944520c
```
stanek-michal: I am not sure if we need to change the ECS ref or not

nfritts (approved):

brian-mckinney (approved):

Makefile
pzl: 1 - yes, this ref will need updating to add these ECS fields
2- CI is not happy about this particular hash:

```
git -C /buildkite/builds/bk-agent-prod-k8s-1783048298343071251/elastic/endpoint-package/out/ecs checkout f04bb41e4fcc35befcaa94a1d9575bc61944520c
fatal: reference is not a tree: f04bb41e4fcc35befcaa94a1d9575bc61944520c
make: *** [Makefile:120: /buildkite/builds/bk-agent-prod-k8s-1783048298343071251/elastic/endpoint-package/out/ecs/.git/refs/tags/f04bb41e4fcc35befcaa94a1d9575bc61944520c] Error 128
```

pzl: ECS updates can often pull in a lot of unrelated changes and balloon a PR (and make it harder to backport). Imagine any updates to field descriptions, limits, other added fields get pulled in, etc.

To keep this small, you may want to undo this ecs ref change, and just duplicate the field description, type, etc and copy it into `custom_schemas/custom_process.yml`.

This should result in the same net change to `package/endpoint/data_stream/process/fields/fields.yml` after running `make`. This is the actual file that gets imported and parsed in kibana to create the mappings, so defining this in `custom_schemas`, to get the same end result, is a-ok.

stanek-michal: done, waiting for CI now

pzl (approved):
yes, this approach will work

you may have to battle with `make` to ensure it generates the changes in the `package/endpoint/docs/README.md` file, add and commit that

but you're good to merge if you can get CI passing

stanek-michal:
Thanks!

PR #764, opened 2026-07-13
[9.6] Add ECS fields for NTDLL enrichment during VirtualProtect/Ex events

ig-commits:
## Change Summary

This PR adds one new fields to `endpoint-package` that are generated by Elastic Defend during VirtualProtect and VirtualProtectEx API enrichment events:

- `process.Ext.api.metadata.target_section`

Proposed schema fields approved in the original endpoint issue by @Samirbous. This PR will support endpoint-dev PR **20987** targeting `9.6`. 

I currently have my `endpoint-dev` branch submodule pointing this branch instead, so I think I am looking to have this PR merged so I can keep my `endpoint-dev` PR pointing at main.

### Sample values

Below is the sample of the data in my testing Kibana instance showing the new json fields present:

```json
    "process": {
      "Ext": {
        "api": {
          "metadata": {
            "target_section": ".text"
          },
```

## Release Target

9.6

### For mapping changes:


pzl:
@ig-commits  are these fields something we want specifically searchable, sortable, filterable in kibana? 

Or are these fields additional context in documents that users will find likely using other fields, but still want to _view_ this information?

If the latter, then these fields don't need explicit mappings. They can be allowlist/whitelist/exceptions if you are validating against EAF or something. The values will always be sent and viewable in kibana. You add mappings when it is a field being used as a search field

0xflux:
Hey @pzl, thanks for the comment. Just FYI I swapped my github account over a few weeks back.

Regarding your comment, the fields will be used for `endpoint-scripts` to search against, likewise for custom detection content from customers should they choose. I can see this being valuable metadata to aid threat hunting also looking through the raw data.

If that doesn't require a change in `endpoint-package`, I'm happy to revert, but as an absolute necessity, we would want `endpoint-scripts` and customer rules to be able to search against it.

pzl:
ok, then do add them here :+1:

pzl (approved):

pzl:
I believe this PR: https://github.com/elastic/endpoint-package/pull/779 should clear the CI blockage for you.

when that gets merged, you can pull in `main`, or rebase, or whatever, and you should be unblocked

0xflux:
@pzl just a heads up, I removed 1 field since your approval in the last commit.  5e4f3138e2bcd8fdd26bdb2e524102c3b753285b

PR #786, opened 2026-08-06
 custom documentation update

intxgo:
## Change Summary

This PR fix two things:

1. Regenerate custom documentation removing stale md files.

2. `make` command never updated custom documentation, which was a source of developer errors, now `make` regenerates the `custom_documentation/doc` tree, also GitHub action will fail if the developer forget to commit regenerated doc files.
 
### Sample values

Sample document:

```json


```

### For mapping changes:

### For Transform changes:


package/endpoint/data_stream/process/fields/fields.yml
```diff
@@ -931,7 +931,7 @@
       level: custom
       type: long
       description: Bitmask of the taint flags this module sets on the kernel. Zero means the module does not taint the kernel.
-      example: 4096
+      example: '4096'
```
pzl: if you pull in `main`, does this change go away now?

Tried to address in #779

pzl (approved):

nfritts (approved):

PR #787, opened 2026-08-07
Add process_vm_access fields for process_vm_readv/writev

Aegrah:
## Summary

Add ECS mappings for Linux `process_vm_access` events so Defend documents can carry
cross-process memory access details from `process_vm_readv` / `process_vm_writev`.
Fields sit under `process.Ext.process_vm_access.*` next to other Linux process telemetry.

Issue: https://github.com/elastic/endpoint-dev/issues/21204

- Data stream/action: process events, `event.action: process_vm_access`
- Fields: `target_pid`, `target_start_time_ns`, `remote_addr`, byte counts and iovec
  counts (`long`); `operation` (`keyword`)
- `target_pid` is resolved by the kernel from the pid passed to the call, so it is a host
  PID even when the caller runs in a PID namespace and is directly comparable to
  `process.pid`. `target_start_time_ns` pairs with it so a target cannot be confused with
  an unrelated process that later reuses the same PID.
- `bytes_transferred` preserves the raw syscall return value, including negative errno
  values for failed attempts
- Generation: sequential `make clean && make all`; only the three
  `process.Ext.process_vm_access.*` fields move

## Proof

- Package generation: PASS
- Release-aligned Fleet acceptance: `20260910T102925Z-process-vm-access` (Endpoint 9.4.2),
  accepted build `endpoint-9.4.2-aba0979dc8d75bb9`
- Persona volume gate: `20260910T114814Z-process-vm-access-volume` on that build
  (PASS, 1h) — **0** `process_vm_access` documents against 91,643 documents/hour
- Accepted documents carried `event.action=process_vm_access` with operation, target
  identity and byte fields

Supersedes the evidence quoted in earlier revisions of this PR: the quark side was
redesigned to resolve the target at `mm_access` rather than from the namespace-relative
syscall argument, and all three gates were rerun against that design.

## Companion PRs

- Quark: https://github.com/elastic/quark/pull/416 (supersedes #376)
- endpoint-dev: https://github.com/elastic/endpoint-dev/pull/21420


Aegrah:
Superseded by https://github.com/elastic/endpoint-package/pull/800. The redesign keeps the hybrid architecture (syscall result + `mm_access` target identity) but replaces the named syscall tracepoints with ABI-aware raw entry/exit hooks (ia32/x32 compat calls were invisible), keeps attempts rejected before target lookup, makes the iovec fields honest bounded snapshots instead of a first-vector "address/size" (a zero-length decoy first vector and a partial transfer inside one vector were both reproduced), fixes the release-branch timestamp/ring-loss accounting, adds the optional-source retry in both Endpoint watchers, and replaces the weak tests (namespace/TID identity, compat entry, permission denial, partial transfer, map pressure, cleanup). Gates: preflight `20260916-pva-v2-release-final` PASS, Fleet acceptance `20260916T172021Z-process-vm-access` PASS (accepted build `endpoint-9.4.2-5564d77d41a8132d`); the persona volume gate was explicitly waived by the owner for this round and is recorded as waived, not passed. Per the workflow rule a redesign opens new PRs instead of force-pushing a different design here; closing this one.

PR #767, opened 2026-07-20
[WIP] Add fields for Linux mprotect events

Aegrah:
## Summary

Add queryable mappings and Linux process-event documentation for:

- `process.Ext.mprotect.addr`
- `process.Ext.mprotect.len`
- `process.Ext.mprotect.prot`
- `process.Ext.mprotect.prot_flags`

The docs describe the final contract: only successful calls requesting executable protection (X, RX, WX, or RWX) are emitted, and collection requires `advanced.events.hidden.emit_diagnostic_fields`.

Issue: https://github.com/elastic/endpoint-dev/issues/20408

## Workflow and proof

The contract workflow regenerated the package and combined schemas, built Endpoint, ran focused unit/EAF tests, deployed the custom integration through Fleet, and asserted Elasticsearch output.

- Preflight `20260722T124741Z-mprotect-preflight`: **PASS**
- Fleet acceptance `20260722T142745Z-mprotect` on Stack/Agent/Endpoint 9.4.2: **PASS**
- `make clean all`, generated handoff, and schema freshness: **PASS**
- Realistic workstation qualification `20260722T103000Z-mprotect-realistic-smoke`: **PASS**

```text
3,708 Endpoint documents / 4,737 action values
exec: 523 (11.04%); mprotect: 6 (0.13%, 0.16% document coverage)
mprotect executables: /usr/bin/node 5; chromium headless_shell 1
masks: RWX 6 (100%)
```

## Companion PRs

- Quark: https://github.com/elastic/quark/pull/367
- endpoint-dev: https://github.com/elastic/endpoint-dev/pull/21136

PR #797, opened 2026-09-08
Add fields for Linux executable mprotect attempt events

nicholasberlin:
## Summary

Add mappings and Linux process-event documentation for executable `mprotect()` attempt events under `process.Ext.mprotect`:

| Field | Type | Meaning |
|---|---|---|
| `vma_start`, `vma_end` | long | Bounds of the VMA the kernel considered (not the address passed to the call) |
| `prev_prot_flags` | keyword | Protection before the call, decoded in fixed R, W, X order; `""` when the area had no permissions |
| `req_prot` | long | Raw protection flags passed to the call |
| `req_prot_flags` | keyword | Requested protection, decoded |
| `effective_prot_flags` | keyword | Kernel-adjusted protection (e.g. `READ_IMPLIES_EXEC`); always contains `X` |
| `file_backed` | boolean | Whether the mapping is backed by a file |
| `file.path`, `file.inode`, `file.device` | keyword | Identity of the backing file, absent for anonymous memory |

The field design follows the final quark implementation (elastic/quark#409, #410, #412): the event is collected at the `security_file_mprotect()` LSM hook, so it describes an executable-memory *attempt* rather than a successful syscall, and quark reports each protection transition once per process life (once per file for file-backed mappings).

Supersedes #767, which described the earlier syscall-tracepoint design.

## Testing

- `make schemas/v1/process/process.yaml` and `go run ./scripts/generate-docs` are clean against the checked-in outputs.
- `package/endpoint/data_stream/process/fields/fields.yml` was hand-edited in the generated style.
- End-to-end coverage (EAF `test_events_mprotect*`) lives in the companion endpoint-dev PR.


stanek-michal:
I thought we weren't supposed to add `Ext` fields anymore and do ECS instead. That said it's pretty low level functionality so idk.

gergoabraham (approved):

pzl (approved):
lgtm

ECS inclusion is a good default goal, but agreed this seems fairly low-level

stanek-michal (approved):

PR #393, opened 2023-07-19
Added Process Rollback fields

bit-envoy:
## Change Summary

Added response fields related to Process Rollback

### Sample values

Sample document:

```json
    "Responses": [
        {
            "@timestamp": "2023-07-19T14:21:05.0Z",
            "action": {
                "action": "process_rollback",
                "process": {
                    "path": "C:\\Users\\Pawel Mirski\\AppData\\Local\\Programs\\Python\\Python38-32\\python.exe"
                }
            },
            "message": "Successful production process rollback",
            "result": 0
        }
    ]
```

### For mapping changes:

- [x] I ran `make` after making the schema changes, and committed all changes
- [x] If these field(s) are "exception"-able, I made a companion PR to Kibana adding it (see [Readme](https://github.com/elastic/endpoint-package#exceptionable))
- [x] If this is a `metadata` change, I also updated both transform destination schemas to match

### For Transform changes:

- [ ] The new transform successfully starts in Kibana
- [ ] The corresponding transform destination schema was updated if necessary

PR #799, opened 2026-09-16
Add Responses.x to for malware and ransomware events

michalpristas:
## Change Summary

This pull request adds support for a new `Responses` field group across malware, memory protection, and ransomware alert schemas and their corresponding custom subset files. These changes introduce a comprehensive set of fields to capture detailed response actions, results, and related metadata for each alert type.

**Schema enhancements for response actions:**

* Added a `Responses` field group to the `malware_event.yaml`, `memory_protection_event.yaml`, and `ransomware_event.yaml` schemas, including fields for action details, timestamps, messages, process information, and result codes. This enables richer tracking and analysis of automated or manual responses taken for each alert. 

**Custom subset updates:**

* Updated the custom subset files `malware_event.yaml`, `memory_protection_event.yaml`, and `ransomware_event.yaml` in the `custom_subsets/elastic_endpoint/alerts/` directory to include the new `Responses` field group, ensuring these fields are available for queries and data extraction. 

### For mapping changes:

### For Transform changes:


tomsonpl (approved):
Nice, thanks for adding these — the generated schemas look consistent with `rule_detection_event.yaml`, and since `Responses.*` is already mapped in the shared `alerts` data stream there's no mapping impact. 👍

Two small questions, neither blocking:

1. **`Responses.process.entity_id` is typed `text`** (in `custom_schemas/custom_responses.yml`) while every other `entity_id` in the package is `keyword`. I know it's pre-existing and not something this PR introduces, but this PR is what makes it visible on malware/memory/ransomware docs — so people will likely try to term-query or pivot on it. Do you know if that was intentional, or is it worth a follow-up to switch it to `keyword`?

2. **Doc field lists** — I noticed `Responses.action.process.message` and `Responses.action.process.result` don't appear in any doc source, and `action.process.path` only shows up under `windows_malware_alert`. Does the sensor emit all three for Process Rollback, or is the current set accurate? Just want to make sure the per-OS splits match reality, since nothing validates these lists.

Thanks!

gergoabraham (approved):
sorry for the delay! 🙌

michalpristas:
@tomsonpl 
re 1 i think it is worth a followup

re 2. it was introduced in https://github.com/elastic/endpoint-package/pull/393/changes 
my bot claims we're not using them, so dont know.
i will probably need to recreate PR to kick off CI i can remove those fields from doc

stanek-michal:
> @tomsonpl re 1 i think it is worth a followup
> 
> re 2. it was introduced in https://github.com/elastic/endpoint-package/pull/393/changes my bot claims we're not using them, so dont know. i will probably need to recreate PR to kick off CI i can remove those fields from doc

looks like we do emit `Responses.action.process.path` in Rollback (but not the other ones)
also noticed `Responses.action.process.path` is only documented for windows malware alert but not for:

  • Windows behavior
  • Windows ransomware
  • Windows shellcode
  • Windows memory-threat (that one doesn't document Responses fields at all but should)

michalpristas:
closing in favor of https://github.com/elastic/endpoint-package/pull/803

PR #803, opened 2026-09-22
Add Responses.x to for malware and ransomware events

michalpristas:
Copy of https://github.com/elastic/endpoint-package/pull/799 but not doing PR from fork to kick off CI

## Change Summary

Adds the existing `Responses` field group to malware, memory protection, and ransomware alert schemas (it was previously only on rule-detection), and documents the fields those alert types actually emit. Also drops two unused Process Rollback fields that were never written by the sensor.

**Schema / mapping**

* Included `Responses: fields: "*"` in the `malware_event`, `memory_protection_event`, and `ransomware_event` custom subsets so those alerts map the same `Responses.*` fields as rule-detection.
* Regenerated `schemas/v1/alerts/{malware,memory_protection,ransomware}_event.yaml`.
* Removed unused `Responses.action.process.message` and `Responses.action.process.result` from `custom_schemas/custom_responses.yml` (leftover from #393). The sensor stores Process Rollback outcome on top-level `Responses.message` / `Responses.result` and only writes `Responses.action.process.path`. Regenerated `rule_detection_event.yaml`, `fields.yml`, and the package README.

**Custom documentation**

* Documented `Responses.{@timestamp,action.action,message,process.*,result}` on malware / memory-threat / ransomware alerts (all OS) and on Windows shellcode-thread.
* Documented `Responses.action.process.path` on Windows malware, ransomware, shellcode-thread, and memory-threat (Process Rollback is Windows-only).

### For mapping changes:

### For Transform changes:


schemas/v1/alerts/malware_event.yaml
```diff
@@ -262,6 +262,229 @@ Endpoint.policy.applied.version:
...
+  dashed_name: Responses-action-key-values-name
+  description: Value name recovered by Rollback
+  flat_name: Responses.action.key.values.name
+  ignore_above: 1024
+  level: custom
+  name: action.key.values.name
+  normalize: []
+  short: Value name recovered by Rollback
+  type: keyword
+Responses.action.process.path:
```
tomsonpl, line 382: Thanks for splitting this out! 🙏 Small question about `325237d` ("removed not needed fields"): it deletes `Responses.action.process.message` and `Responses.action.process.result` from around here, but `custom_schemas/custom_responses.yml` still defines both (lines 99 and 104) and this subset asks for `Responses: fields: "*"`.

`schemas/v1/alerts/rule_detection_event.yaml` uses the same directive and still has both fields (lines 146 and 166), so I think `make` would put these 19 lines back — leaving the next person with a confusing unrelated diff. Doesn't look like CI would catch it either; `check.sh` only runs `make static-test` and `make pipeline-test`.

Was the intent to drop these two for real? If so, would taking them out of `custom_responses.yml` work instead — though that also removes them from `rule_detection_event.yaml` and the package mapping, so maybe worth a quick check with the team? Or if they should stay, is reverting that commit simplest? 🙂

schemas/v1/alerts/memory_protection_event.yaml
```diff
@@ -289,6 +289,229 @@ Memory_protection.unique_key_v2:
...
+  dashed_name: Responses-action-key-values-name
+  description: Value name recovered by Rollback
+  flat_name: Responses.action.key.values.name
+  ignore_above: 1024
+  level: custom
+  name: action.key.values.name
+  normalize: []
+  short: Value name recovered by Rollback
+  type: keyword
+Responses.action.process.path:
```
tomsonpl, line 409: Same two fields removed here as in `malware_event.yaml` — same question, just noting it applies to all three generated files. 🙂

schemas/v1/alerts/ransomware_event.yaml
```diff
@@ -548,6 +548,229 @@ Ransomware.version:
...
+  dashed_name: Responses-action-key-values-name
+  description: Value name recovered by Rollback
+  flat_name: Responses.action.key.values.name
+  ignore_above: 1024
+  level: custom
+  name: action.key.values.name
+  normalize: []
+  short: Value name recovered by Rollback
+  type: keyword
+Responses.action.process.path:
```
tomsonpl, line 668: And the third one — same as above. 🙂

custom_documentation/src/endpoint/data_stream/alerts/windows/windows_ransomware_alert.yaml
```diff
@@ -61,6 +61,9 @@ fields:
   - Responses.action.source.attributes
   - Responses.action.source.path
   - Responses.message
+  - Responses.process.entity_id
```
tomsonpl, line 65: Thanks for adding these! Just flagging (not re-litigating, since I think it was already raised on #799): `Responses.action.process.path` is documented only for Windows malware, while ransomware/shellcode/memory-threat emit it too. Happy to leave as-is if that was already settled over there.

custom_documentation/src/endpoint/data_stream/alerts/windows/windows_memory_threat_alert.yaml
```diff
@@ -29,6 +29,13 @@ fields:
   - Endpoint.policy.applied.id
   - Endpoint.policy.applied.name
   - Endpoint.policy.applied.version
+  - Responses.@timestamp
```
tomsonpl, line 32: Also carried over from #799 I think: this one gets 7 `Responses` fields while `windows_shellcode_thread` gets 16, even though both resolve to `memory_protection_event`. Is that intentional (memory_signature genuinely never triggers file/registry rollback)? No strong opinion, just curious. 🙂

tomsonpl (approved):
Thanks for splitting this out! 🙏 Small question about `325237d` ("removed not needed fields"): it deletes `Responses.action.process.message` and `Responses.action.process.result` from around here, but `custom_schemas/custom_responses.yml` still defines both (lines 99 and 104) and this subset asks for `Responses: fields: "*"`.

`schemas/v1/alerts/rule_detection_event.yaml` uses the same directive and still has both fields (lines 146 and 166), so I think `make` would put these 19 lines back — leaving the next person with a confusing unrelated diff. Doesn't look like CI would catch it either; `check.sh` only runs `make static-test` and `make pipeline-test`.

Was the intent to drop these two for real? If so, would taking them out of `custom_responses.yml` work instead — though that also removes them from `rule_detection_event.yaml` and the package mapping, so maybe worth a quick check with the team? Or if they should stay, is reverting that commit simplest? 🙂

schemas/v1/alerts/malware_event.yaml
michalpristas: yeah i removed it from my diff, `schemas/v1/alerts/rule_detection_event.yaml` contained them previously, i'm not sure if they were used but quick scan tells me they are not used now. 
i could remove them but i wanted this PR to be adding missing pieces, i can do followup with cleanup

custom_documentation/src/endpoint/data_stream/alerts/windows/windows_ransomware_alert.yaml
michalpristas: i will add them, this makes sense to be in this PR

schemas/v1/alerts/malware_event.yaml
gergoabraham: @michalpristas please make sure to include those removals in the PR description as well

gergoabraham (approved):

brian-mckinney (approved):

